PR #14, opened 2022-08-12
[CI] Enable CI and setup all logistics

adam2392:
First step for #11 

#### Summary for CI

We should keep CI and setup and installation workflow relatively similar. From my exp in OSS, this alleviates developer burden and overall technical debt.

This PR adds additional checks to the CI that are based off best practices:

- circleCI will build documentation for every PR and commit to this repo
- GH actions has a job to link output of circleCI to be linked to the checks
- GH actions will test for style using poetry
- GH actions will test for building the overall package
- GH actions will run unit tests
- GH actions will upload code coverage from unit tests
- GH actions will also automatically release on pypi given a tag is made on `main` branch

TODO for CI:
- GH actions should also weekly test integration suite. I.e. this will eventually encompass networkx, dowhy, pywhy_graphs and more.
- ideally poetry doesn't install all dependencies for every single job it runs

#### Summary for repo docs

I've added more details to the README based on my experience. What would be nice is to extend the main figure in dowhy's README to encompass causal discovery. Otw, I think short READMEs are the way to go and then linking to the documentation.

#### Summary for code documentation

I've uploaded boiler plate documentation that will leverage the pydata sphinx layout. Inside we should track: contributors, and PRs. The API is also auto-rendered here via sphinx.

cc: @robertness


.github/workflows/main.yml
```diff
@@ -0,0 +1,198 @@
...
+        with:
+          poetry-version: ${{ matrix.poetry-version }}
+
+      - name: Install Poetry Dynamic Versioning Plugin
+        run: pip install poetry-dynamic-versioning
+
+      # TODO: once poetry v1.2 is release, we want to only install dev dependencies
+      - name: Install dependencies
+        run: poetry install
+
```
adam2392, line 45: Note @darthtrevino there is no way that I am aware of to only install what's needed for running the style checks. This ends up installing all dependencies right?

I think there is a fix coming up in v1.2 for poetry, but for now, I suppose this is fine since reading more, poetry does have a lot of nice properties.

PR #16, opened 2022-08-16
[ENH] CI tests for constraint-based structure learning

adam2392:

Follow up to: #14 to lead to #11

- Adds conditional independence testing to the submodule `dodiscover.ci` in a consistent API.
- Also removed `pylint` from the style checking as it is too stringent.

Needs to update the networkx algorithms that will work on mixedEdgeGraphs.

## Before submitting

## After submitting


pyproject.toml
```diff
@@ -61,35 +60,6 @@ graphviz = { version = "^0.20.1" }
 dowhy = { version = "^0.8" }
 ipython = { version = "^7.4.0" }
 nbsphinx = { version = "^0.8" }
-# [tool.poetry.extras]
```
darthtrevino, line 64: I see, so the sphinx-related dev-dependencies are always going to install in CI. That's fine - in dowhy we did that slightly differently but this should hold until poetry 1.2.

dodiscover/ci/_script.py
```diff
@@ -0,0 +1,63 @@
...
+from dodiscover.ci import g_square_binary, g_square_discrete
+from dodiscover.ci.tests import testdata
+
+if __name__ == "__main__":
+
+    dm = np.array([testdata.bin_data]).reshape((5000, 5))
+    x = 0
+    y = 1
+
+    sets = [[], [2], [2, 3], [3, 4], [2, 3, 4]]
```
darthtrevino, line 16: Should this test script be put under the `tests/` folder? Is this something we expect users to run? It seems more likely that they'd use the conditional-independence classes.

dodiscover/ci/fisher_z_test.py
```diff
@@ -0,0 +1,116 @@
...
+    sample_size = data.shape[0]
+    var = list({x, y}.union(sep_set))  # type: ignore
+    (var_idx,) = np.in1d(data.columns, var).nonzero()
+
+    # compute the correlation matrix within the specified data
+    sub_corr_matrix = correlation_matrix[np.ix_(var_idx, var_idx)]
+    # try:
+    inv = np.linalg.inv(sub_corr_matrix)
+    r = -inv[0, 1] / sqrt(inv[0, 0] * inv[1, 1])
+    # except Exception as e:
```
darthtrevino, line 101: We should probably delete this commented-out block

darthtrevino (approved):

robertness:
Looking at this now.

Makefile
```diff
@@ -0,0 +1,5 @@
+build-docs:
+	@echo "Building documentation"
+	make -C doc/ clean
+	make -C doc/ html-noplot
+	cd doc/ && make view
```
robertness, line 5: Newline.
Does the makefile need to be commited?

adam2392: No. This is just for my dev sake rn. I'm still wrangling upstream dependencies. I.e. pywhy-graphs. I'll remove it.

adam2392:
I'll get these CI green and then work on the PC part. 

This turned out to be a lot of work since we're trying to keep integration with networkx proposal, and pywhy-graphs. But I think it'll be worth it once we see the PC algo implemented with these in mind :).

dodiscover/_protocol.py
```diff
@@ -0,0 +1,52 @@
+from typing import Dict, FrozenSet, Iterable, Protocol
+
+import networkx as nx
+
+
+class GraphProtocol(Protocol):
```
robertness, line 6: Interesting.  What is the rationale for using duck typing instead of subclassing?

dodiscover/ci/_script.py
robertness: And if this is a test script, should it not use unittest or some alternative and assert some outcome is correct?

dodiscover/ci/_script.py
```diff
@@ -0,0 +1,63 @@
...
+    for idx in range(len(sets)):
+        print("x =", x, ", y =", y, ", s =", sets[idx], end="")
+        _, p = g_square_binary(dm, x, y, set(sets[idx]))
+        print(", p =", p, end="")
+        fr_p = frexp(p)
```
robertness, line 21: I assume this code is breaking down the p value float into components so you can assert approximate equality.  Is this necessary?  Is this a place where we would see a lot of numeric variation across devices?

dodiscover/ci/base.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def _check_test_input(self, df: pd.DataFrame, x_var, y_var, z_covariates):
+        if any(col not in df.columns for col in [x_var, y_var]):
+            raise ValueError("The x and y variables are not both in the DataFrame.")
+        if z_covariates is not None and any(col not in df.columns for col in z_covariates):
+            raise ValueError("The z conditioning set variables are not all in the DataFrame.")
+
+    @abstractmethod
+    def test(
+        self, df: pd.DataFrame, x_var: Any, y_var: Any, z_covariates: Any = None
```
robertness, line 23: Why are `x_var`, `y_var` and `z_covariates` of the `Any` type?  Categoricals, integers, and floats right?

dodiscover/ci/fisher_z_test.py
```diff
@@ -0,0 +1,107 @@
...
+        Parameters
+        ----------
+        correlation_matrix : np.ndarray of shape (n_variables, n_variables), optional
+            ``None`` means without the parameter of correlation matrix and
+            the correlation will be computed from the data., by default None
+        """
+        self.correlation_matrix = correlation_matrix
+
+    def test(
+        self, df: pd.DataFrame, x_var: Any, y_var: Any, z_covariates: Any = None
```
robertness, line 24: Instead of Any these should be continuous right?

dodiscover/ci/base.py
adam2392: They can be anything that a pandas dataframe column supports. I wasn't 100% sure if `Categoricals, integers, and floats` are sufficient. Are they?

dodiscover/_protocol.py
adam2392: Protocols just are Python's way of defining an abstract interface. So in the code, note that `pywhy-graphs` is never imported. And does not need to be. This is in line w/ discussions we had awhile back where the Dev team requested that we don't explicitly have to rely on pywhy's implementation of causal graphs.

This `GraphProtocol` class then allows a user to define their own graph class, as long as it adheres to the API, then it "should" still work. In addition, this provides `mypy` type checking w/o having to import `pywhy-graphs` explicitly.

dodiscover/ci/_script.py
adam2392: Ah yeah. I think this can be deleted tbh. It was some debugging code.

dodiscover/_protocol.py
adam2392: Just to add an explicit example:

if someone from say `ananke` wants to make their graph implementation work with the implementation of the algorithms here that rely on a GraphProtocol, then they just need to modify their API to have the functions specified.

I personally am okay w/ being opinionated and requiring users to use pywhy-graphs directly and providing a transformation module for converting graphs from common formats, but I think we can always swap that easily using ctrl+f+replace

dodiscover/ci/fisher_z_test.py
```diff
@@ -0,0 +1,107 @@
...
+    ) -> Tuple[float, float]:
```
robertness, line 25: Might be useful to create types for the test stat and the p-value as this will be used in multiple places.

adam2392: Aren't they just floats tho? With pvalue being constrained between 0 and 1

Makefile
adam2392: Actually jk I think it would be nice to have this. It makes the docs building a bit easier. If someone knows how to add a multi-step procedure to the poetry command `build_docs`, then maybe not tho...

darthtrevino: If it's something scripted like this I would generally model this under a bash file and invoke the bash file via a poetry script.

dodiscover/ci/g_test.py
```diff
@@ -0,0 +1,396 @@
...
+    Parameters
+    ----------
+    x : int | str
+        the first node variable. If ``data`` is a DataFrame, then
+        'x' must be in the columns of ``data``.
+    y : int | str
+        the second node variable. If ``data`` is a DataFrame, then
+        'y' must be in the columns of ``data``.
+    sep_set : set
+        the set of neibouring nodes of x and y (as a set()).
```
robertness, line 38: typo "neibouring".

dodiscover/ci/oracle.py
```diff
@@ -0,0 +1,66 @@
...
+    ----------
+    graph : nx.DiGraph | pywhy_graphs.ADMG
+        The ground-truth causal graph.
+    """
+
+    def __init__(self, graph: GraphProtocol) -> None:
+        self.graph = graph
+
+    def test(self, df, x_var, y_var, z_covariates):
+        """Conditional independence test given an oracle.
```
robertness, line 23: Did you want to add a reference?

tests/unit_tests/ci/test_fisherz_test.py
```diff
@@ -0,0 +1,27 @@
...
+
+from dodiscover.ci import FisherZCITest
+
+seed = 12345
+
+
+def test_fisher_z():
+    """Test Fisher Z test for Gaussian data."""
+    rng = np.random.RandomState(seed)
+    X = rng.randn(300, 1)
```
robertness, line 12: Docstrings here that explain what is expected

robertness (approved):
LGTM.  My only issues are with some of the type hints.

@adam2392 can you comment here on why some flavors are left out?  For example:

- mutual information and asymptotic chi-square test for discrete variables, perhaps with a James-Stein estimator
- Monte Carlo permutation testing and sequential Monte Carlo permutation testing
- Jonckheere-Terpstra test for ordinal variables

adam2392:
Will address the code comments tonight. 

Re the additional CI tests: just never implemented them myself :p. It would be nice to get a community contribution for those. Or one of us can tackle it when there's time?

tests/unit_tests/ci/test_fisherz_test.py
adam2392: Done.

dodiscover/ci/oracle.py
adam2392: Done.

adam2392:
Gonna merge this for now to move forward w/ the next step. Left some comments open w/ my responses tho.

PR #18, opened 2022-08-18
[ENH] Add PC algorithm

adam2392:
Fixes #11
Closes: #10 

Changes proposed in this pull request:

- PC algorithm implementation with separate skeleton discovery stage (has all the features of causal-learn I think(?), except for missing-value support)
- Adds the abstract class for Constraint Based learning algorithms
- Adds the `Context` class
- Adds an example showing how to run PC algorithm
- Adds a metric submodule for comparing graphs

## Before submitting

- [ ] I've read and followed all steps in the [Making a pull request](https://github.com/py-why/dodiscover/blob/main/CONTRIBUTING.md#making-a-pull-request)
    section of the `CONTRIBUTING` docs.
- [ ] I've updated or added any relevant docstrings following the syntax described in the
    [Writing docstrings](https://github.com/py-why/dodiscover/blob/main/CONTRIBUTING.md#writing-docstrings) section of the `CONTRIBUTING` docs.
- [ ] If this PR fixes a bug, I've added a test that will fail without my fix.
- [ ] If this PR adds a new feature, I've added tests that sufficiently cover my new functionality.

## After submitting

- [ ] All GitHub Actions jobs for my pull request have passed.

PR #20, opened 2022-08-19
[ENH] Skeleton learning method

adam2392:
Follow on to: #19 
And part of #18 

-

## Before submitting

## After submitting


dodiscover/constraint/skeleton.py
```diff
@@ -0,0 +1,470 @@
...
+import networkx as nx
+import numpy as np
+
+from dodiscover.ci import BaseConditionalIndependenceTest
+
+from ..context import Context
+
+logger = logging.getLogger()
+
+ACCEPTED_MARKOVIAN_SKELETON_METHODS = ["complete", "neighbors", "neighbors_path"]
```
bloebp, line 15: What about defining this as enums? Then we can avoid the usual issues with string parameters and the options are easier to discover.

dodiscover/constraint/skeleton.py
```diff
@@ -0,0 +1,470 @@
...
+
+
+def _iter_conditioning_set(
+    possible_variables: Iterable,
+    x_var: Any,
+    y_var: Any,
+    size_cond_set: int,
+):
```
bloebp, line 23: I noted that there no return types (here or at other places). Maybe you can add them as well, e.g., ```-> List[Any]```.

dodiscover/constraint/skeleton.py
```diff
@@ -0,0 +1,470 @@
...
+
+logger = logging.getLogger()
+
+ACCEPTED_MARKOVIAN_SKELETON_METHODS = ["complete", "neighbors", "neighbors_path"]
+
+
+def _iter_conditioning_set(
+    possible_variables: Iterable,
+    x_var: Any,
+    y_var: Any,
```
bloebp, line 21: Would it be possible to specific the types here (and at the other places in the code) further? E.g., as numpy array or union of numpy array, pandas dataframe etc.

dodiscover/constraint/skeleton.py
```diff
@@ -0,0 +1,470 @@
...
+
+
+def _find_neighbors_along_path(G: nx.Graph, start, end) -> Set:
+    def _assign_weight(u, v, edge_attr):
+        if u == node or v == node:
+            return np.inf
+        else:
+            return 1
+
+    nghbrs = set()
```
bloebp, line 62: Just a minor thing, but sometimes it might be better to be more descriptive about the parameter names. Here, instead of ```nghbrs``` simply write it out ```neighbors```. Its not that much longer, but looks much less cryptic.

dodiscover/constraint/skeleton.py
```diff
@@ -0,0 +1,470 @@
...
+    remove_edges: Set
+
+    def __init__(
+        self,
+        ci_estimator: BaseConditionalIndependenceTest,
+        sep_set: Dict[str, Dict[str, List[Set[Any]]]] = None,
+        alpha: float = 0.05,
+        min_cond_set_size: int = 0,
+        max_cond_set_size: int = None,
+        max_combinations: int = None,
```
bloebp, line 182: For ```None``` parameters, add the ```Optional``` type. For instance, ```max_cond_set_size: Optional[int] = None```.

dodiscover/constraint/skeleton.py
```diff
@@ -0,0 +1,470 @@
...
+
+        # store the list of potential adjacencies for every node
+        # which is tracked and updated in the algorithm
+        adjacency_mapping: Dict[Any, List] = dict()
+        for node in nodes:
+            adjacency_mapping[node] = [
+                other_node for other_node in adj_graph.neighbors(node) if other_node != node
+            ]
+
+        logger.info(
```
bloebp, line 278: There should be an option to turn off logging information. Maybe make the logger an optional parameter that can be passed or add flag to turn it off. (This applies to all logging/print information)

dodiscover/constraint/skeleton.py
```diff
@@ -0,0 +1,470 @@
...
+            )  # type: ignore
+
+        if x_var in possible_variables:
+            possible_variables.remove(x_var)
+        if y_var in possible_variables:
+            possible_variables.remove(y_var)
+
+        return possible_variables
+
+    def _postprocess_ci_test(
```
bloebp, line 450: Since the 'classical PC-algorithm' doesn't do an adjust to keep the FDR, would it be possible to extend this in the future? E.g., something like https://www.jmlr.org/papers/volume10/li09a/li09a.pdf.

dodiscover/constraint/skeleton.py
```diff
@@ -0,0 +1,470 @@
...
+        self, adj_graph: nx.Graph, x_var, y_var, cond_set: Set, test_stat: float, pvalue: float
+    ) -> bool:
+        # keep track of the smallest test statistic, meaning the highest pvalue
+        # meaning the "most" independent
+        if np.abs(test_stat) <= adj_graph.edges[x_var, y_var]["test_stat_abs"]:
```
bloebp, line 455: Wouldn't it be enough to store ```test_stat``` instead of ```test_stat_abs```? If the absolute value is needed, one can simply add an ```abs(..)```.

dodiscover/constraint/utils.py
```diff
@@ -0,0 +1,51 @@
+from typing import Any, Dict, List, Set
+
+import pandas as pd
+
+
+def dummy_sample(G):
```
bloebp, line 6: Type of ```G``` is missing.

dodiscover/constraint/utils.py
```diff
@@ -0,0 +1,51 @@
...
+    """Sample an empty dataframe with columns as the nodes.
+
+    Used for oracle testing.
+    """
+    df_values = dict()
+    for node in G.nodes:
+        df_values[node] = []
+
+    df = pd.DataFrame.from_dict(df_values)
```
bloebp, line 15: - Could be reduced to ```return pd.DataFrame({column: [] for column in G.nodes})```.
- Note that ```node``` could potentially have a data type that is incompatible with the requirements for pandas column names.

dodiscover/context.py
```diff
@@ -0,0 +1,114 @@
...
+    is used in conjunction with a discovery algorithm.
+
+    Setting the apriori explicit direction of an edge is not supported yet.
+    """
+
+    def __init__(
+        self,
+        data: pd.DataFrame,
+        variables: Optional[Set] = None,
+        latents: Optional[Set] = None,
```
bloebp, line 50: How would we use the ```latents``` information here? Would it be sufficient to simply indicate between which variables we expect latent confounders?

tests/unit_tests/constraint/test_skeleton.py
```diff
@@ -0,0 +1,119 @@
...
+                    1: (2, 3),
+                    2: (),
+                    3: (),
+                    4: (3,),
+                }
+            ),
+        ),
+    ],
+)
+def test_learn_skeleton_with_data(indep_test_func, data_matrix, g_answer, alpha=0.01):
```
bloebp, line 90: Is there a reason for ```alpha = 0.01``` instead of the 'common' ```0.05```?

dodiscover/constraint/skeleton.py
```diff
@@ -0,0 +1,470 @@
...
+        if self.min_cond_set_size is None:
+            self.min_cond_set_size_ = 0
+        else:
+            self.min_cond_set_size_ = self.min_cond_set_size
+        if self.max_combinations is None:
+            self.max_combinations_ = np.inf
+        else:
+            self.max_combinations_ = self.max_combinations
+
+    def fit(self, context: Context):
```
bloebp, line 239: This only returns the skeleton, but does not direct edges when possible (using the 4 rules, such as colliders)? Maybe it makes sense to add a function to direct the edges as well based on the separation sets.

tests/unit_tests/constraint/test_skeleton.py
adam2392: Just cuz the CI test gets the wrong answer at 0.05. I added a comment.

dodiscover/context.py
adam2392: Just following the API here: https://github.com/py-why/dowhy/wiki/API-for-Global-Causal-Discovery

I think latents would not normally be used unless maybe doing some type of sensitivity analysis where you assume all of a sudden "x" is unobserved

dodiscover/constraint/utils.py
adam2392: Good point, I'll maybe add an error message if not number or string. If we run into errors later by users, I'm sure we can revisit the point.

dodiscover/constraint/skeleton.py
adam2392: WDYT about using networkx partners? so instead something like `nbr` so it's shorter and not prone to misspelling?

ref: https://github.com/networkx/networkx/pull/5898#discussion_r937274643

dodiscover/constraint/skeleton.py
adam2392: These would be anything that can be a data frame column, so not numpy array, or union of numpy array. I can just constrain the type then for now to what makes sense for us: `str` and `numerics`.

dodiscover/constraint/skeleton.py
adam2392: Added `Iterable[Set]`

dodiscover/constraint/skeleton.py
adam2392: Added the enumeration. Lmk wyt.

dodiscover/constraint/skeleton.py
adam2392: Doesn't logging allow the user to set "logging level" to say "error" and they can automatically not see any of the info logs?

Or do you mean just encompass `logger.info` messages perhaps in some type of `verbose` conditional check?

dodiscover/constraint/skeleton.py
adam2392: Sure that makes sense.

dodiscover/constraint/skeleton.py
adam2392: Yes definitely. I tried to tease apart every single part of the algorithm as much as possible, such that any stage can be extended in arbitrary ways. E.g. sorting by dependencies, cutting off the algorithm for high number of combinations of variables, etc.

Should we start an issue to add this in? I think some basic FDR would be nice feature

adam2392:
This now gives us a generic backbone for all the PC/FCI algo variants, where the first step is "learning a skeleton" from CI testing of the data.

The `LearnSkeleton` can most likely be subclassed for additional structure such as computing the "possibly d-separating sets" in RFCI/FCI and evaluation of F-nodes when given interventional data.

dodiscover/constraint/skeleton.py
bloebp: Looks good. You can now also directly use it as a type in the code then, i.e. ```skeleton_method: SkeletonMethods``` and do the comparison using the enums ````if self.skeleton_method == SkeletonMethods.COMPLETE```.

Btw. consider using capital letters for the enum values.

dodiscover/constraint/skeleton.py
bloebp: This would be the data type within the data frame or numpy array. But here we don't ask for a single float or str, but an array/dataframe. So I was rather thinking of something like
```
x_var: Union[np.ndarray, pd.DataFrame],
y_var: Union[np.ndarray, pd.DataFrame],
```
We don't further specify the data type within the array/frames.

dodiscover/constraint/skeleton.py
bloebp: I don't have so many experience with the logger, but since you initiate a new logger at the beginning, I am not sure if this would honor your settings (or if the settings are global). Maybe worth to simply try it out :)

dodiscover/constraint/skeleton.py
bloebp: Yea sounds good. I also never saw a proper FDR version of it, so this would be definitely something new.

dodiscover/constraint/skeleton.py
adam2392: See: #25 for the fix. If that is acceptable, I'll add those here too.

dodiscover/constraint/skeleton.py
adam2392: Tracking this now in #24

tests/unit_tests/constraint/test_skeleton.py
```diff
@@ -0,0 +1,121 @@
...
+                    4: (3,),
+                }
+            ),
+        ),
+    ],
+)
+def test_learn_skeleton_with_data(indep_test_func, data_matrix, g_answer):
+    """Test PC algorithm for estimating the causal DAG."""
+    # the tests get the correct independences when alpha is at 0.01
+    alpha = 0.01
```
bloebp, line 93: Hmm, 'tuning' the parameter in way to make the test work is maybe not a good idea. There might be something wrong with the implementation, which is then not visible due to the tune parameter. If this is purely due to the flakiness of the (statistical) tests, consider rerunning them them when they fail with newly generated data. If it fails, e.g., 3x in a row, then it is a good indication to double check the implementation (and this could be the independence test here). See for instance https://github.com/py-why/dowhy/blob/main/tests/gcm/independence_test/test_kernel.py#L12, where we use flaky for exactly these kind of flaky scenarios.

adam2392: IIRC I think I took the data from pcalg of the R package... so it's not actually generated on the fly.

I assumed it was fine since the tests passed other types of simulated data, but let me actually investigate. 

I might just play around w/ simulating some new data instead from an SCM.

adam2392: Oh wait I just checked again and 0.05 works. I think this might've been a different private project that I was working on. The alpha level works for 0.05 (default) and 0.01.

dodiscover/constraint/skeleton.py
adam2392: Added the `Column` type.

bloebp (approved):

Issue #17, opened 2022-08-18
Add additional CI tests: Mutual Information, Chi square, Monte Carlo


adam2392:
LGTM.  My only issues are with some of the type hints.

@adam2392 can you comment here on why some flavors are left out?  For example:

- mutual information and asymptotic chi-square test for discrete variables, perhaps with a James-Stein estimator
- Monte Carlo permutation testing and sequential Monte Carlo permutation testing
- Jonckheere-Terpstra test for ordinal variables

_Originally posted by @robertness in https://github.com/py-why/dodiscover/pull/16#pullrequestreview-1076586407_

adam2392:
Other tests that I'm trying to use for some projects are:

- CCIT: https://arxiv.org/pdf/1709.06138.pdf
- CMI: https://proceedings.mlr.press/v84/runge18a/runge18a.pdf
- CCMI: https://arxiv.org/abs/1906.01824

adam2392:
Additional info on the Jonckheree test: 

https://www.quantics.co.uk/blog/jonckheere-terpstra-test-ordered-differences/
https://www.stats.ox.ac.uk/~steffen/teaching/fsmHT07/fsm07.pdf
https://en.wikipedia.org/wiki/Jonckheere%27s_trend_test

C implementation in R: https://github.com/cran/bnlearn/blob/master/src/tests/discrete/jonckheere.c

Issue #26, opened 2022-08-24
Tracking CI tests and moving them to a new repo?


adam2392:
Most likely we will be moving the CI tests from dodiscover/dowhy to a new CI testing repo. Then this thread talks about some of the design choices of type of input to expect.

_Originally posted by @adam2392 in https://github.com/py-why/dodiscover/pull/25#discussion_r953085358_

adam2392:
All CI tests in https://github.com/py-why/dowhy/tree/96a055ae4448ecddeb458d86b10588ea3295c4a3/dowhy/gcm/independence_test should also get transferred. 

Most likely, this should occur after we do an initial release of pywhy-graphs and dodiscover v0.1, so we can then refactor out the CI tests and make a separate package release of `pycit` v0.1 (i.e. Python Conditional Independence Testing).

adam2392:
TODO:
- [ ] Add `GeneralizedCovCITest`
- [ ] Add `ApproxKernelCITest` (using the RCIT approach)
- [ ] `RegressionCITest`

Note: all unit tests should also get migrated as well. https://github.com/py-why/dowhy/tree/96a055ae4448ecddeb458d86b10588ea3295c4a3/tests/gcm/independence_test

Once this is done, I think the next step is to migrate this out of both repos, so that way development can occur faster, independently and the whole documentation can be catered towards discussing pros/cons and "gotchas" of CI testing. This can then become an optional dependency in dodiscover and dowhy. 

This helps remove LOC from dodiscover and dowhy, helps remove dependencies in dowhy and can consolidate a new package that is solely dedicated to maintaining a consistent and robust API for (conditional) independence testing

adam2392:
Moving to pywhy-stats

PR #28, opened 2022-08-24
[ENH] Add the classifier CI test

adam2392:
Addresses one of: #17 

- Implements the CCIT proposed test which is a nonparametric conditional independence test that allows multivariate X, Y and Z
- Adds a simulation function from the CCIT paper
- The nearest-neighbor sampling procedure proposed to generate samples from the null hypothesis: $X \perp Y | Z$ is possibly of independent interest beyond just the CCIt function. I remember seeing the same algorithm in the [CCMI](https://arxiv.org/abs/1906.01824) paper too, which I am interested in for my projects. I will prolly PR that sometime soon too.

A note on class/function design for CI tests:
- I'm more of a fan of the API for class-based CI test vs function-based. This helps demonstrate that. There are so many parameters for a sklearn classifier, and ideally a user can instantiate that separately from the CI estimator parameters. 

## Before submitting

## After submitting


bloebp:
> I'm more of a fan of the API for class-based CI test vs function-based. This helps demonstrate that. There are so many parameters for a sklearn classifier, and ideally a user can instantiate that separately from the CI estimator parameters.

There might be a misconception with the idea of the functional form. There is no need to ask for model parameters in the CI function. For instance:

```
def my_ci_test(X, Y, Z, ml_model):
   ...
   ml_model.fit(...)

mdl = MyComplexMLModel(**millions_of_parametrs)

p_value = my_ci_test(X, Y, Z, mdl)
```

You can also expect a factory instead that generates a user specific ML model instead of expecting the parameter as inputs to the function.

dodiscover/ci/clf_test.py
```diff
@@ -0,0 +1,359 @@
...
+from numpy.typing import NDArray
+from sklearn.neighbors import NearestNeighbors
+from sklearn.utils import check_random_state
+
+from dodiscover.typing import Column
+
+from .base import BaseConditionalIndependenceTest
+
+
+class ClassifierCITest(BaseConditionalIndependenceTest):
```
bloebp, line 17: Seems very closely related to https://arxiv.org/abs/1804.02747, which is also implemented (in a significantly improved form) here: https://github.com/py-why/dowhy/blob/main/dowhy/gcm/independence_test/regression.py#L14

A general problem with these ML prediction based approaches is, that they don't find all kind of dependencies, only when they impact the expectation. For instance, a dependency with respect to the variance:

```
import numpy as np

from dowhy.gcm import kernel_based, regression_based

X = np.random.normal(0, 1, 1000)
Y = []

for x in X:
    Y.append(np.random.normal(0, abs(x)))

Y = np.array(Y)
Z = np.random.normal(0, 1, 1000)

print("Correct result:", kernel_based(X, Y, Z))
print("Wrong result", regression_based(X, Y, Z))

clf = RandomForestClassifier()
ci_estimator = ClassifierCITest(clf)

df = pd.DataFrame({'x': X, 'y': Y, 'z': Z})

_, pvalue = ci_estimator.test(df, {"x"}, {"z"}, {"y"})
print("Wrong result", pvalue)
```

Here, ```X``` and ```Y``` are dependent (even after conditioning on ```Z```, which is irrelevant here), but only the kernel based test would find this dependence, while the ML based test would claim they are independent.

Nevertheless, it is still worth having it implemented :)

dodiscover/ci/clf_test.py
```diff
@@ -0,0 +1,359 @@
...
+    def __init__(
+        self,
+        clf: sklearn.base.BaseEstimator,
+        metric: Callable = sklearn.metrics.accuracy_score,
+        bootstrap: bool = False,
+        n_iter: int = 20,
+        correct_bias: bool = True,
+        threshold: float = 0.03,
+        test_size: Union[int, float] = 0.3,
+        random_state: Optional[int] = None,
```
bloebp, line 27: The random state should be set outside of the class objects. If a user wants to fix the seed, simply set it before calling the code.

petergtz:
> Ultimately, you would only gain something from the class-based approach when you plan to reuse the object and want to save some passing of parameters (which is of course a valid argument). However, we should make sure that these tests are stateless then.

Agree with @bloebp here. Classes make sense when something requires a longer life-span, e.g. because you want to pass it around and it gets invoked at a different place than where it's constructed. This is typically also the case in bigger systems where you want to separate wiring of objects from invoking them. Another case is, when something requires multiple steps to build, such as a graph object.

But in scenarios where we instantiate a class and then immediately invoke the object as suggested by the unit test:

```python
    ci_estimator = ClassifierCITest(clf, random_state=rng)

    _, pvalue = ci_estimator.test(df, {"x"}, {"x1"})
    assert pvalue > 0.05
    _, pvalue = ci_estimator.test(df, {"x"}, {"z"})
    assert pvalue < 0.05
```
the benefit of a class seems really questionable. Seems like this could be written as:

```python
    _, pvalue = classifier_ci_test(df, {"x"}, {"x1"}, clf)
    assert pvalue > 0.05
    _, pvalue = classifier_ci_test(df, {"x"}, {"z"}, clf)
    assert pvalue < 0.05
```

And this would save a lot of bookkeeping code in the implementation too.

What do you think?

adam2392:
> And this would save a lot of bookkeeping code in the implementation too.
> 
> What do you think?

Okay this makes a lot of sense! Let me start another issue to track the refactoring of the CI tests into functions. Also tagging #26 since we discussed moving these altogether from do discover + dowhy to another repo.

dodiscover/ci/clf_test.py
adam2392: Internally, I wanted the bootstrapping / random sampling to be consistent across runs, so the result is the same no matter what. If I only set a seed, then the following give difff results:

```
np.random.seed(12345)
stat, pvalue =ccit(x,y,z, )
```
vs.
```
np.random.seed(12345)
do_something_before()
stat, pvalue =ccit(x,y,z, )
```

https://stackoverflow.com/questions/56119505/what-is-the-difference-between-np-random-seed-vs-random-state

dodiscover/ci/clf_test.py
adam2392: This is a great point. I'll note it in the docs and provide an example showing the counter-example. I guess I can do that in this PR.

dodiscover/ci/clf_test.py
adam2392: But if you run multiple random functions in sequence then you would have to re seed the global seed before every function to make it 100% reproducible? My impression was that setting a global seed in general is undesirable (Tho not sure of the detailed intricacies besides this issue).

I just adopted a sklearn pattern here https://scikit-learn.org/stable/developers/utilities.html

adam2392: Ah okay I think I found why: https://stackoverflow.com/questions/59105921/why-is-numpy-random-seed-not-remaining-fixed-but-randomstate-is-when-run-in-para

IIUC the global seed is fine for now but we would in principle add joblib to every function. At that point I think explicitly calling np.random.* seems to have some issues that numpy warns against.

bloebp: We also use global seeds with joblib. Its correct that this requires a careful implementation of the parallelization, but take for instance a look at:
- https://github.com/py-why/dowhy/blob/main/dowhy/gcm/independence_test/kernel.py#L111
- https://github.com/py-why/dowhy/blob/main/dowhy/gcm/independence_test/kernel.py#L487
- https://github.com/py-why/dowhy/blob/main/dowhy/gcm/shapley.py#L219

where we solve this simply by generating a list of random seeds (based on the current global random state) and pass these seeds into the parallel processes. Since the generated seeds are deterministic with a certain global seed, the parallel runs are deterministic as well. This allows you to have also a deterministic behavior when using joblib.

dodiscover/ci/clf_test.py
robertness: > A general problem with these ML prediction based approaches is, that they don't find all kind of dependencies, only when they impact the expectation. For instance, a dependency with respect to the variance:

I don't suppose this is a huge problem for dependence induced by _causality_.  I'm trying to think of an example of an SCM where a causal parent would only affect the variance of the child.  Maybe a if X and Y were parents of Z, Z is a sensor reading, X is the true signal, and Y is a source of interference in the sensor reading?

dodiscover/ci/clf_test.py
```diff
@@ -0,0 +1,359 @@
...
+
+from dodiscover.typing import Column
+
+from .base import BaseConditionalIndependenceTest
+
+
+class ClassifierCITest(BaseConditionalIndependenceTest):
+    def __init__(
+        self,
+        clf: sklearn.base.BaseEstimator,
```
robertness, line 20: Can we make this compatible with a deep learning based classifier?

robertness (changes requested):
Could we expand this method to work with a neural-net based classifier using Keras or Pytorch?  It likely feels like overkill but it would be a first step to telling a story to the broader community that this repo is using cutting edge "deep" methods.  We could also workshop an example with multidimensional variables later (e.g. pixels) though not in this PR.

dodiscover/ci/clf_test.py
adam2392: I changed the type and also added it in the docstring. I think adding an example would definitely be helpful. I'm not as familiar with the torch/keras/tensorlfow API as I am with scikit-learn. This could be a good example for a new contributor.

I started an issue.

adam2392:
I've addressed the comments and now just need a PR approval to move forward.

robertness (dismissed):
I think this is at a good place and we can get it merged.  If we need to make changes later we can.

adam2392:
@robertness can you re-approve this PR? I had to merge in changes from main and resolve conflicts, which dismissed your approval. 

Not sure why tho... it seems that's slightly redundant, since it's fairly often that one might need to rebase/merge in changes from main. I guess it's to protect changes from conflicting.

robertness (approved):
LGTM

PR #30, opened 2022-08-26
[ENH] PC algorithm with updated example using dowhy.gcm for the SCM module

adam2392:
Fixes #18 
Closes: #10 

- This supersedes #18
- This is a "big PR", but the example goes hand-in-hand w/ the algorithm implementation. In general, I think any PR with "algorithms" should require at least a corresponding example, if not corresponding example + in-depth tutorial

## Before submitting

## After submitting

### DIFF
--- a/dodiscover/constraint/pcalg.py
+++ b/dodiscover/constraint/pcalg.py
@@ -1,0 +1,309 @@
+import logging
+from itertools import combinations, permutations
+from typing import Dict, List, Optional, Set
+
+import networkx as nx
+
+from dodiscover.ci.base import BaseConditionalIndependenceTest
+from dodiscover.constraint.utils import is_in_sep_set
+from dodiscover.typing import Column
+
+from .._protocol import EquivalenceClassProtocol
+from ._classes import BaseConstraintDiscovery
+
+logger = logging.getLogger()
+
+
+class PC(BaseConstraintDiscovery):
+    """Peter and Clarke (PC) algorithm for causal discovery.
+
+    Assumes causal sufficiency, that is, all confounders in the
+    causal graph are observed variables. See :footcite:`Spirtes1993` for
+    full details on the algorithm.
+
+    Parameters
+    ----------
+    ci_estimator : Callable
+        The conditional independence test function. The arguments of the estimator should
+        be data, node, node to compare, conditioning set of nodes, and any additional
+        keyword arguments.
+    alpha : float, optional
+        The significance level for the conditional independence test, by default 0.05.
+    min_cond_set_size : int, optional
+        Minimum size of the conditioning set, by default None, which will be set to '0'.
+        Used to constrain the computation spent on the algorithm.
+    max_cond_set_size : int, optional
+        Maximum size of the conditioning set, by default None. Used to limit
+        the computation spent on the algorithm.
+    max_iter : int
+        The maximum number of iterations through the graph to apply
+        orientation rules.
+    max_combinations : int, optional
+        Maximum number of tries with a conditioning set chosen from the set of possible
+        parents still, by default None. If None, then will not be used. If set, then
+        the conditioning set will be chosen lexographically based on the sorted
+        test statistic values of 'ith Pa(X) -> X', for each possible parent node of 'X'.
+    apply_orientations : bool
+        Whether or not to apply orientation rules given the learned skeleton graph
+        and separating set per pair of variables. If ``True`` (default), will
+        apply Meek's orientation rules R0-3, orienting colliders and certain
+        arrowheads :footcite:`Meek1995`.
+    ci_estimator_kwargs : dict
+        Keyword arguments for the ``ci_estimator`` function.
+
+    Attributes
+    ----------
+    graph_ : EquivalenceClassProtocol
+        The equivalence class of graphs discovered.
+    separating_sets_ : dict of dict of list of set
+        The dictionary of separating sets, where it is a nested dictionary from
+        the variable name to the variable it is being compared to the set of
+        variables in the graph that separate the two.
+
+    References
+    ----------
+    .. footbibliography::
+    """
+
+    graph_: EquivalenceClassProtocol
+    separating_sets_: Optional[Dict[Column, Dict[Column, List[Set[Column]]]]]
+
+    def __init__(
+        self,
+        ci_estimator: BaseConditionalIndependenceTest,
+        alpha: float = 0.05,
+        min_cond_set_size: int = None,
+        max_cond_set_size: int = None,
+        max_iter: int = 1000,
+        max_combinations: int = None,
+        apply_orientations: bool = True,
+        **ci_estimator_kwargs,
+    ):
+        super().__init__(
+            ci_estimator,
+            alpha,
+            min_cond_set_size=min_cond_set_size,
+            max_cond_set_size=max_cond_set_size,
+            max_combinations=max_combinations,
+            **ci_estimator_kwargs,
+        )
+        self.max_iter = max_iter
+        self.apply_orientations = apply_orientations
+
+    def convert_skeleton_graph(self, graph: nx.Graph) -> EquivalenceClassProtocol:
+        """Convert skeleton graph as undirected networkx Graph to CPDAG.
+
+        Parameters
+        ----------
+        graph : nx.Graph
+            Converts a skeleton graph to the representation needed
+            for PC algorithm, a CPDAG.
+
+        Returns
+        -------
+        graph : EquivalenceClassProtocol
+            The CPDAG class.
+        """
+        from pywhy_graphs import CPDAG
+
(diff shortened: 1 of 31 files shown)


dodiscover/constraint/pcalg.py
```diff
@@ -0,0 +1,309 @@
...
+class PC(BaseConstraintDiscovery):
+    """Peter and Clarke (PC) algorithm for causal discovery.
+
+    Assumes causal sufficiency, that is, all confounders in the
+    causal graph are observed variables. See :footcite:`Spirtes1993` for
+    full details on the algorithm.
+
+    Parameters
+    ----------
+    ci_estimator : Callable
```
robertness, line 26: Does the fact that this is a callable leave us open to new methods in CI estimation?
If so, then great.
I think the ability to write custom CI tests using new methods is a powerful way to utilize cutting-edge research.

dodiscover/constraint/pcalg.py
```diff
@@ -0,0 +1,309 @@
...
+
+        # convert Graph object to a CPDAG object with
+        # all undirected edges
+        graph = CPDAG(incoming_undirected_edges=graph)
+        return graph
+
+    def orient_edges(self, graph: EquivalenceClassProtocol) -> None:
+        """Orient edges in a skeleton graph to estimate the causal DAG, or CPDAG.
+
+        These are known as the Meek rules :footcite:`Meek1995`. They are deterministic
```
robertness, line 117: How would these work with required edges and excluded edges?  Does it make sense to build in those constraints now?  Those are the only real domain knowledge constraints we can provide for constraint-based models.

dodiscover/constraint/utils.py
```diff
@@ -14,7 +15,11 @@ def dummy_sample(G: GraphProtocol):
 
 
 def is_in_sep_set(
-    check_var, sep_set: Dict[str, Dict[str, List[Set[Any]]]], x_var, y_var, mode: str = "any"
+    check_var,
+    sep_set: Dict[Column, Dict[Column, List[Set[Column]]]],
```
robertness, line 19: Please create a custom sep-set type.  This will improve readability.

dodiscover/context.py
```diff
@@ -112,3 +120,21 @@ def excluded_edges(self) -> nx.Graph:
...
+    def copy(self):
+        """Create a copy of the Context object.
+
+        Returns
+        -------
+        context : Context
+            A copy.
+        """
+        context = Context(
+            data=copy(self._data),
```
robertness, line 133: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.copy.html

dodiscover/context.py
```diff
@@ -112,3 +120,21 @@ def excluded_edges(self) -> nx.Graph:
...
+            variables=copy(self._variables),
+            latents=copy(self._latents),
+            init_graph=self._init_graph.copy(),
```
robertness, line 136: This will produce a shallow copy of the graph.  Would need `copy.deepcopy` otherwise

dodiscover/context.py
```diff
@@ -112,3 +120,21 @@ def excluded_edges(self) -> nx.Graph:
     @property
     def init_graph(self) -> GraphProtocol:
         return self._init_graph
+
+    def copy(self):
```
robertness, line 124: I expect, given our current planning for context, that a copy method on a context object should produce [deep copies](https://docs.python.org/3/library/copy.html#copy.deepcopy). ping @darthtrevino

dodiscover/metrics.py
```diff
@@ -0,0 +1,66 @@
+import networkx as nx
```
robertness, line 1: Definitely could have been its own PR :)

dodiscover/metrics.py
```diff
@@ -0,0 +1,66 @@
...
+import numpy as np
+from sklearn.metrics import confusion_matrix
+from sklearn.preprocessing import LabelBinarizer
+
+from ._protocol import GraphProtocol
+
+
+def confusion_matrix_networks(
```
robertness, line 9: FYI: 
* [issue on SID](https://github.com/py-why/dodiscover/issues/41)
*[issue on SHD](https://github.com/py-why/dodiscover/issues/40)
* [issue on visualizing graph comparison](https://github.com/py-why/dodiscover/issues/39)

examples/plot_pc_alg.py
```diff
@@ -0,0 +1,184 @@
+"""
```
robertness, line 1: Make this a jupyter notebook.

tests/unit_tests/constraint/test_pcalg.py
```diff
@@ -0,0 +1,255 @@
...
+
+from dodiscover import Context
+from dodiscover.ci import GSquareCITest, Oracle
+from dodiscover.constraint import PC
+from dodiscover.constraint.utils import dummy_sample
+from dodiscover.metrics import confusion_matrix_networks
+from dodiscover.testdata.testdata import bin_data, dis_data
+
+
+@pytest.mark.parametrize(
```
robertness, line 16: @darthtrevino any thoughts on this testing approach, relative to unittest?

tests/unit_tests/constraint/test_pcalg.py
```diff
@@ -0,0 +1,255 @@
...
+    assert added_arrows
+    assert G.has_edge("C", "A", G.directed_edge_name)
+
+    # if 'B' and 'D' are adjacent, then it will
+    # not be added
+    G = G_copy.copy()
+    G.add_edge("B", "D", G.directed_edge_name)
+    added_arrows = pc._apply_meek_rule3(G, "C", "A")
+    assert not added_arrows
+
```
robertness, line 216: Can we include some tests that show the expected resolution when the Meek rules collide with fixed edges?

robertness (approved):
I'm okay with this so long as my requests are addressed.

Also, can you double check that you have tests in place that will protect this code against pending refactors to the context object?

dodiscover/constraint/pcalg.py
adam2392: Yep! It is any Callable with the `BaseConditionalIndependenceTest` API. However, I will end up refactoring all those to just "functions" as suggested by @bloebp and @petergtz. In a downstream PR this will look like the following:

- ctrl+f / replace `BaseConditionalIndependenceTest` with `Callable[[np.ndarray, np.ndarray, np.ndarray], float]` for type hints
- just turn all the CI test classes into their corresponding functions
- migrate any internal functionality of the CI test classes into private util functions for the `dodiscover/ci/` submodule

adam2392: Basically this will be very powerful because as long as CI tests adhere to a specific input/output function signature, then yes this enables anyone to arbitrarily pass in whatever CI test they want. This is an API pro over say passing in "strings", where you only allow certain CI tests.

dodiscover/context.py
adam2392: Changed to deepcopy

dodiscover/context.py
adam2392: Changed to deepcopy. Yeah this is a prototype just to demonstrate how it would work. Most likely will be superseded by the API in #42

tests/unit_tests/constraint/test_pcalg.py
adam2392: I'm not following what you mean here. Can you elaborate @robertness ?

tests/unit_tests/constraint/test_pcalg.py
adam2392: pytest is significantly better from my exp :p

adam2392:
> Also, can you double check that you have tests in place that will protect this code against pending refactors to the context object?

The context object is only accessed through its class properties rn and then the instantiations are used in the unit tests. They are used in conjunction w/ the PC algo, so in principle changes to the context that break PC algo, will be detected by those tests.

I've addressed most of your comments, except https://github.com/py-why/dodiscover/pull/30#discussion_r964201595. In addition, some misc. topics:

- The inclusion/exclusion of edges I think is a larger issue, which I've begun a thread in #46. 
- Another lower-pri item is the refactoring of CI tests into functions #45

robertness (dismissed):
I think we should get this merged if there are no further concerns

tests/unit_tests/constraint/test_pcalg.py
robertness: I'm just wondering how included and excluded edges, for example, will work with these rules.

robertness (approved):
I agree that we can address include/exclude edge issues elsewhere.

adam2392:
> I agree that we can address include/exclude edge issues elsewhere.

Sorry can you re-approve? Fixing the merge conflicts inside changelog and conf.py and it disables your previous approval :(

robertness (approved):
Reapproving

Issue #29, opened 2022-08-25
Implement GES using causal-learn


robertness:
**Is your feature request related to a problem? Please describe.**
We need an initial score-based algorithm.

**Describe the solution you'd like**
Use causal-learn as a dependency to build GES

**Describe alternatives you've considered**
We could build it on our own, but causal-learn has a solid implementation and there is no need to reinvent the wheel

adam2392:
Here is a rendered doc that is a WIP of the PC algorithm API and its usage: https://output.circle-artifacts.com/output/job/4d057804-45d2-488b-89de-11fd8d37743c/artifacts/0/dev/index.html

This base class also defines the interface for any "constraint" causal discovery algorithm: https://github.com/py-why/dodiscover/blob/ab79e40853438ea6c5918554627b55a343c47ef2/dodiscover/constraint/_classes.py. Note this might change as Robert is updating the docs around `Context` vs `data`.

adam2392:
From the call, it seems the action items are:

- [ ] implement a notebook/script that calls causal-learn and gets GES working on some example
- [ ] write a function/class in dodiscover that wraps causal-learn's GES function and implements an api
- [ ] add causal-learn to the `pyproject.toml` dev dependencies (I can help you do this)
- [ ] write some unit/integration tests

There prolly will need to be some iterative discussion on the GES API. Basically, my general intuition is that all score-based algorithms should subclass a base class, or have a standard function signature.

#### Some thoughts on GES wrapping causal-learn

https://causal-learn.readthedocs.io/en/latest/search_methods_index/Score-based%20causal%20discovery%20methods/GES.html contains the parametrization of GES in causal-learn. Note the differences relative to what dodiscover does. 

1. data is a numpy array rather than dataframe
2. parameters imo should be "snake cased", like 'max_parents' instead of 'maxP'
3. score function: should it be a `Callable` [object](https://www.geeksforgeeks.org/callable-in-python/), or str? If callable, then what is the expected signature of any score function?  For example, see [sklearn's metrics](https://scikit-learn.org/stable/modules/model_evaluation.html#scoring-parameter), which all have a standard function signature of `y_true, y_pred, *, normalize=True, sample_weight=None`.

Lmk if I missed anything?

PR #32, opened 2022-08-30
[ENH; WIP] Implementation of FCI algorithm that leverages the base classes used in PC

adam2392:

- implements the FCI algorithm
- implements a subclass of `LearnSkeleton` for learning skeleton graphs using the idea of "possibly-d-separating" sets and "PDS-Path" sets (see RFCI paper)
- implements unit tests for skeleton and FCI procedure

Next PR:
- [ ] implements an example demonstrating how FCI is different

This relies on Pywhy-graph changes in: https://github.com/py-why/pywhy-graphs/pull/10
Also this will be downstream of #30 .

Order of merging:
1. #30 
2. this PR

## Before submitting

## After submitting

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,7 +67,7 @@ jobs:
             sudo apt install libspatialindex-dev xdg-utils
       - python/install-packages:
           pkg-manager: poetry
-          args: "-E graph_func -E viz"
+          args: "-E graph_func -E viz --with docs"
           cache-version: "v1" # change to clear cache
       - run:
           name: Check poetry package versions
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,13 +5,15 @@ concurrency:
 on:
   pull_request:
     paths:
-      - '**.py'
+      - "**.py"
+      - pyproject.toml
+      - .github/workflows/main.yml
   push:
     branches: [main]
     paths:
-      - '**.py'
+      - "**.py"
     tags:
-      - 'v*.*.*'
+      - "v*.*.*"
   workflow_dispatch:
 
 jobs:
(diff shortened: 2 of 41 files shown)


bloebp:
Just want to mention that if we start supporting causal-learn (which we plan to), we should probably not have another implementation of the most common algorithms. However, I see the benefit here to demonstrate the API and graph usages.

adam2392:
> Just want to mention that if we start supporting causal-learn (which we plan to), we should probably not have another implementation of the most common algorithms. However, I see the benefit here to demonstrate the API and graph usages.

That sounds good. I'm motivated to implement at least the core constraint-based algos. in this repo and codebase because of a few reasons:

i) extensibility: rn I am trying to get an implementation of rFCI working w/o having to use the R package pcalg.
ii) maintainability: I don't see the rules 5-10 of Zhang 2008 in the [FCI function](https://github.com/cmu-phil/causal-learn/blob/main/causallearn/search/ConstraintBased/FCI.py). I found it easier to just draft up an implementation that teases apart the semantics

In the proposed [constraint-based API](https://github.com/py-why/dodiscover/blob/main/dodiscover/constraint/_classes.py), I think it's important to modularize and pull apart each aspect of the core algorithms to enable future improvements. E.g. rFCI, or even conservativeFCI, or maxvoteFCI is trivially implementable by subclassing `class FCI` here. 

Of course... this all relies on me being invested in constraint-based learning :p. I think if there is developer consensus on the API, then re-implementing will be helpful in the long run to consolidate the API across the entire pywhy. With that being said, I am less motivated to do the score-based algorithms, which I think then implementing a wrapper sounds like a great way of getting the functionality out right away. E.g. #29

PR #35, opened 2022-08-31
[FIX] Rename Protocol classes

### LINKED ISSUE #33: Proposal: rename GraphProtocol to Graph
Hi @adam2392,

I recommend renaming

https://github.com/py-why/dodiscover/blob/3c4f8d251dcf080c4bce83292affff6cd46f335d/dodiscover/_protocol.py#L6

to just `Graph`. In Java it became kind of an anti-pattern to name an entity `ISomeEntity` for interfaces and I would apply the same here. The reason is that it gives you less flexibility in your design to change from a concrete class to a protocol or back, because you'd have to change the name too and everywhere where it's used.

adam2392:
Fixes #33 

- Updated GraphProtocol -> Protocol
- Updated unit tests to reflect changes in pywhy-graphs

## Before submitting

## After submitting


bloebp:
Is there a reason to not be explicit about that they are 'Protocols'?

adam2392:
> Is there a reason to not be explicit about that they are 'Protocols'?

Isn't that what was suggested here tho? https://github.com/py-why/dodiscover/issues/33

petergtz:
> Is there a reason to not be explicit about that they are 'Protocols'?

@bloebp Have a look at https://developer.okta.com/blog/2019/06/25/iinterface-considered-harmful and replace Java interfaces with Python Protocols and the `I`-prefix with the `Protocol`-suffix.

TL;DR: As a user of `Graph`, you don't care that it's a Protocol. You just care that about its methods:

```python
def f(g: Graph):
    g.nodes() ...

f(nx.DiGraph())
```

petergtz (dismissed):
Assuming you'll squash commits before merging.

petergtz (approved):

PR #37, opened 2022-09-05
Remove license in doc/index.rst

petergtz:
BSD is not the right license and in another part of this file the correct one is listed by linking to the LICENSE.md file, which avoids the redundance.


robertness (approved):
Thanks for catching this.

PR #42, opened 2022-09-07
[ENH, REFACTOR] Use a builder pattern for context construction

darthtrevino:
@robertness  - This PR is playing around with the "Context Builder" pattern we've been batting around. 

## Before submitting

- [X] If this PR adds a new feature, I've added tests that sufficiently cover my new functionality.

## After submitting

- [x] All GitHub Actions jobs for my pull request have passed.


adam2392:
I like the API. It actually seems reminiscent of the latest changes in [seaborn](https://seaborn.pydata.org/tutorial/objects_interface.html)

darthtrevino:
@adam2392 I think this is okay now, but the pytest invocations are failing in CI - I think it's failing in main too though.

dodiscover/context_builder.py
```diff
@@ -0,0 +1,140 @@
...
+    _graph: Optional[Graph] = None
+    _included_edges: Optional[Union[nx.Graph, nx.DiGraph]] = None
+    _excluded_edges: Optional[Union[nx.Graph, nx.DiGraph]] = None
+    _observed_variables: Optional[Set[str]] = None
+    _latent_variables: Optional[Set[str]] = None
+
+    def __init__(self, data: pd.DataFrame) -> None:
+        self._data = data
+
+    def data(self, data: pd.DataFrame) -> "ContextBuilder":
```
robertness, line 30: Let's not include data in context.

dodiscover/context_builder.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+
+    _graph: Optional[Graph] = None
+    _included_edges: Optional[Union[nx.Graph, nx.DiGraph]] = None
+    _excluded_edges: Optional[Union[nx.Graph, nx.DiGraph]] = None
+    _observed_variables: Optional[Set[str]] = None
+    _latent_variables: Optional[Set[str]] = None
+
+    def __init__(self, data: pd.DataFrame) -> None:
+        self._data = data
```
robertness, line 28: Let's not use data in context.  @adam2392 and @petergtz and others have convinced me to separate these out.  Others you and I have talked to recently @darthtrevino have convinced me as well.

robertness: I acknowledge that separating data from context might impact @adam2392 's work but I think it would be a mild change.

dodiscover/context_builder.py
```diff
@@ -0,0 +1,140 @@
...
+            nor ``variables`` is set, then it is presumed that ``variables`` consists
+            of the columns of ``data`` and ``latents`` is the empty set.
+
+        Returns
+        -------
+        ContextBuilder
+            The builder instance
+        """
+        self._observed_variables = observed_variables
+        self._latent_variables = latent_variables
```
robertness, line 106: @adam2392 how cool is this?

dodiscover/context_builder.py
```diff
@@ -0,0 +1,140 @@
...
+        Returns
+        -------
+        ContextBuilder
+            The builder instance
+        """
+        self._included_edges = included_edges
+        self._excluded_edges = excluded_edges
+        return self
+
+    def features(
```
robertness, line 84: I think we should say "variables."  This library will also take on some causal representation learning, and in that setting causal variables and "features" are different (features will be a lower level of abstraction.

dodiscover/context_builder.py
```diff
@@ -0,0 +1,140 @@
...
+        A dataset in data-frame form, consisting of samples as rows and variables as columns.
+    """
+
+    _graph: Optional[Graph] = None
+    _included_edges: Optional[Union[nx.Graph, nx.DiGraph]] = None
+    _excluded_edges: Optional[Union[nx.Graph, nx.DiGraph]] = None
+    _observed_variables: Optional[Set[str]] = None
+    _latent_variables: Optional[Set[str]] = None
+
+    def __init__(self, data: pd.DataFrame) -> None:
```
robertness, line 27: Let's have the constructor use nodes instead of data.

dodiscover/context_builder.py
robertness: If we need something in the constructor, I vote for this

robertness (approved):
Approved with requested changes re data

dodiscover/context_builder.py
adam2392: So the new API for algos should be: `algo.fit(data, context)`

?

dodiscover/context_builder.py
```diff
@@ -0,0 +1,140 @@
...
+    Parameters
+    ----------
+    data : pd.DataFrame
+        A dataset in data-frame form, consisting of samples as rows and variables as columns.
+    """
+
+    _graph: Optional[Graph] = None
+    _included_edges: Optional[Union[nx.Graph, nx.DiGraph]] = None
+    _excluded_edges: Optional[Union[nx.Graph, nx.DiGraph]] = None
+    _observed_variables: Optional[Set[str]] = None
```
adam2392, line 24:
```suggestion
    _observed_variables: Optional[Set[Column]] = None
```

In `dodiscover.typing.py`, there is a definition for the typing of variables that we can store.

dodiscover/context_builder.py
```diff
@@ -0,0 +1,140 @@
...
+    _latent_variables: Optional[Set[str]] = None
```
adam2392, line 25: Same here for using `Column` type.

dodiscover/context_builder.py
```diff
@@ -0,0 +1,140 @@
...
+            data=self._data,
+            init_graph=self._graph,
+            included_edges=self._included_edges,
+            excluded_edges=self._excluded_edges,
+            variables=self._observed_variables,
+            latents=self._latent_variables,
+        )
+
+
+def make_context(data: pd.DataFrame) -> ContextBuilder:
```
adam2392, line 127: What would `make_context` do if we don't use data inside Context @robertness ?

adam2392 (changes requested):
Seems like there's some discussion to be had about certain API discussions?

dodiscover/context_builder.py
robertness: @adam2392 the constructor will require a set of nodes for the graph.  This would include nodes that are not included in observed data.

dodiscover/context_builder.py
darthtrevino: @adam2392 yep, latest changes reflect this

dodiscover/context_builder.py
darthtrevino: @robertness  I've updated the builder so that the `data` parameter is optional on the `variables` method, which will enable variable inference. If this is not desired we can rethink this

doc/whats_new/_contributors.rst
```diff
@@ -21,3 +21,4 @@
 
 
 .. _Adam Li: https://pywhy.github.io
```
adam2392, line 23:
```suggestion
.. _Adam Li: https://py-why.github.io
```

doc/whats_new/_contributors.rst
```diff
@@ -21,3 +21,4 @@
...
+.. _Chris Trevino: https://pywhy.github.io
```
adam2392, line 24:
```suggestion
.. _Chris Trevino: https://py-why.github.io
```

adam2392 (approved):
LGTM except the links!

dodiscover/context.py
```diff
@@ -127,9 +98,13 @@ def excluded_edges(self) -> nx.Graph:
...
+    @property
+    def observed_variables(self) -> Set[Column]:
+        return self._variables
+
+    @property
+    def latent_variables(self) -> Set[Column]:
+        return self._latents
 
     def add_state_variable(self, name: str, var: Any) -> None:
         """Add a state variable.
```
adam2392, line 110: Is this part okay? 

So the motivation is that during the `LearnSemiMarkovianSkeleton` algorithm, you need to store an additional PAG, which is the "initial skeleton" so that way you can enumerate the PDS sets per node. Think of it as storing a temporary skeleton to query neighbors smartly.

Would it be better to add that capability into the Builder?

cc: @darthtrevino @robertness

adam2392: Otw, I think what would need to occur is inside the `LearnSemiMarkovianSkeleton`, one would need to re-call the builder and build the same Context, but add additional variables (i.e. the intermediate PAG/skeleton)

See here in FCI: https://github.com/py-why/dodiscover/pull/52/files#r981754091

adam2392: Lmk if that needs clarification.

darthtrevino: Should be okay, but the builder should have a state-variable map it can be initialized with... wip

dodiscover/context_builder.py
```diff
@@ -0,0 +1,214 @@
...
+    """Create a new ContextBuilder instance.
+
+    Returns
+    -------
+    ContextBuilder
+        The new ContextBuilder instance
+    """
+    result = ContextBuilder()
+    if context is not None:
+        result.graph(context.init_graph)
```
adam2392, line 210: Could we just make use of context.copy() here to get all necessary info? 

Just thinking that these LOC might become a pain to maintain every time we add something to context.

adam2392:
Huh weird that those unit tests are failing. I'll take a look.

adam2392 (approved):
LGTM!

adam2392:
@robertness if this is good to you, feel free to squash and merge!

robertness (approved):
LGTM.  I will circle back to the variable interpolation mechanism as we discussed.

PR #47, opened 2022-09-13
use poetry for publication

darthtrevino:
Signed-off-by: Chris Trevino <chtrevin@microsoft.com>

Fixes #

- Use poetry 1.2 in main.yml
- Use poetry to install dynamic versioning plugin. This is the recommended method for 1.2.0+
- Use `poetry publish` in release task

## Before submitting

- [X] I've read and followed all steps in the [Making a pull request](https://github.com/py-why/dodiscover/blob/main/CONTRIBUTING.md#making-a-pull-request)

## After submitting

- [X] All GitHub Actions jobs for my pull request have passed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -150,7 +150,7 @@ jobs:
       #   run: |
       #     pip uninstall -yq networkx
       #     pip install --progress-bar off git+https://github.com/networkx/networkx
-      - name: Run pytest # headless via Xvfb on linux
+      - name: Run pytest
         run: poetry run poe unit_test
       - name: Upload coverage stats to codecov
         if: ${{ matrix.os == 'ubuntu' && matrix.python-version == '3.10' && matrix.networkx == 'stable' }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -121,7 +121,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, macos, windows]
-        python-version: [3.8, "3.10"] # oldest and newest supported versions
+        python-version: [3.8, "3.10"]  # oldest and newest supported versions
         poetry-version: [1.2.0]
         networkx: [stable, main]
     name: pytest ${{ matrix.os }} - py${{ matrix.python-version }} - Networkx ${{ matrix.networkx }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,7 +37,7 @@ jobs:
           poetry-version: ${{ matrix.poetry-version }}
 
       - name: Install Poetry Dynamic Versioning Plugin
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -91,8 +91,8 @@ graph_func = ['pywhy-graphs']
 viz = ['pygraphviz']
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
```
adam2392, line 95: What's the motivation for this change?

adam2392 (approved):
LGTM!

adam2392 (changes requested):
Seems CI is running into some errors.

pyproject.toml
darthtrevino: According to [their docs](https://github.com/mtkennerly/poetry-dynamic-versioning), this is the recommended configuration for the dynamic-versioning plugin in 1.2.0

darthtrevino: It looks like swapping the build-backend caused the sdist checker to fail. Reverting to pip-based install.

.github/workflows/main.yml
```diff
@@ -145,7 +150,7 @@ jobs:
       #   run: |
       #     pip uninstall -yq networkx
       #     pip install --progress-bar off git+https://github.com/networkx/networkx
-      - name: Run pytest  # headless via Xvfb on linux
+      - name: Run pytest # headless via Xvfb on linux
```
adam2392, line 153:
```suggestion
      - name: Run pytest
```

doc/api.rst
```diff
@@ -37,8 +37,8 @@ Graph protocols
 .. autosummary::
    :toctree: generated/
 
-   GraphProtocol
-   EquivalenceClassProtocol
+   Graph
+   EquivalenceClass
```
adam2392, line 41: I think the DCO rebasing makes these PR diffs nonsensical for some reason. E.g. this diff isn't technically part of this PR, but the previous PR... rip. idk how to fix (usually it is to rebase, but seeing as how you did..)

adam2392 (changes requested):
Hey @darthtrevino since you're a more experienced dev, can we try to fix the diff since it seems to show the diff from #35 too for some reason...

- Perhaps `git pull main`
- `git rebase ci/poetry_publish main` ?

If it doesn't work... then I guess we'll just have to deal w/ the DCO weirdness that rebases things weirdly?

The code itself LGTM. The build GH action seems like something that we'll have to test when we slate the first release, where someone needs to upload their PYPI admin username/pw to the GH secrets. But that can be saved for a later date... Easy enough fix.

adam2392 (approved):
I think the diff thing, we can handle in #42...

PR #48, opened 2022-09-20
[ENH] Add metrics module for comparing networks-like graphs

adam2392:
Please review this PR first before #30. cc @robertness 

- Implements `confusion_matrix_networks` function for comparing networkx-like graphs for evaluating causal discovery algorithms
- Update poetry files

Is part of #30 for the sake of lowering LOC diff.

## Before submitting

## After submitting


dodiscover/metrics.py
```diff
@@ -0,0 +1,66 @@
...
+    See Also
+    --------
+    sklearn.metrics.confusion_matrix
+    """
+    if set(true_graph.nodes) != set(pred_graph.nodes):
+        raise RuntimeError("Both nodes should match.")
+
+    # convert graphs to adjacency graph in networkx
+    true_graph = true_graph.to_undirected()
+    pred_graph = pred_graph.to_undirected()
```
bloebp, line 39: Why to do we only want to look at undirected edges? The correctness of the direction could also be measured?

dodiscover/metrics.py
```diff
@@ -0,0 +1,66 @@
...
+from ._protocol import Graph
+
+
+def confusion_matrix_networks(
+    true_graph: Graph,
+    pred_graph: Graph,
+):
+    """Compute the confusion matrix comparing a predicted graph from the true graph.
+
+    Converts the graphs into adjacency matrices, which are symmetric.
```
bloebp, line 15: A adjacency matrix doesn't need to be symmetric, e.g., when it represents a directed graph.

dodiscover/metrics.py
adam2392: This function is meant solely to test correctness of adjacencies which is done commonly in addition to directionality. Whereas something like SHD would handle the case of directionality too. 

I can update the doc string to elaborate.

bloebp: Ok, but note that there (seems to be) multiple definitions of a "adjacency matrix". Here, you assume that `a_{i, j} = a_{j, i}` which does not necessarily need to be the case for a "adjacency matrix".

dodiscover/metrics.py
robertness: I don't think the confusion matrix is symmetric in general?  Is it not the following?

|TP: True Edges ∩ Detected Edges | FN: True Edges ∩ Undetected Edges|
|FP: False Edges ∩ Detected Edges | TN: False Edges ∩ Undetected Edges|

dodiscover/metrics.py
```diff
@@ -0,0 +1,66 @@
...
+
+    See Also
+    --------
+    sklearn.metrics.confusion_matrix
+    """
+    if set(true_graph.nodes) != set(pred_graph.nodes):
+        raise RuntimeError("Both nodes should match.")
+
+    # convert graphs to adjacency graph in networkx
+    true_graph = true_graph.to_undirected()
```
robertness, line 38: So we're comparing undirected edges.  Perhaps this should be an option?

dodiscover/metrics.py
```diff
@@ -0,0 +1,66 @@
...
+    pred_graph = pred_graph.to_undirected()
+
+    # get the order of the nodes
+    idx = np.argsort(true_graph.nodes)
+    other_idx = np.argsort(pred_graph.nodes)
+
+    # next convert into 2D numpy array format and make sure nodes are ordered accordingly
```
robertness, line 45: I like this annotation.  Thank you.

dodiscover/metrics.py
```diff
@@ -0,0 +1,66 @@
...
+    true_adj_mat = nx.to_numpy_array(true_graph)[np.ix_(idx, idx)]
+    pred_adj_mat = nx.to_numpy_array(pred_graph)[np.ix_(other_idx, other_idx)]
+
+    # ensure we are looking at symmetric graphs
+    true_adj_mat += true_adj_mat.T
```
robertness, line 50: I don't understand why the adjacency matrix for an undirected graph would ever not be symmetric.

dodiscover/metrics.py
bloebp: Btw. I was referring to the adjacency matrix here (not confusion matrix).

dodiscover/metrics.py
adam2392: The confusion matrix is not well defined for directed edges, so there would need to be some discussion on how to best extend this in that direction. The SHD is a better approach imo for evaluating directionality of edges as well. This function would be used primarily to evaluate Skeleton learning methods.

dodiscover/metrics.py
adam2392: Yep, good point. Removed this redundancy.

dodiscover/metrics.py
adam2392: To elaborate: how to penalize the incorrect directed edge vs an undirected edge? Is it a -1, -2?

dodiscover/metrics.py
adam2392: I have updated the docstring.

dodiscover/metrics.py
adam2392: I have updated the docstring.

adam2392:
I have updated the docstring to elaborate that this function is solely used to compare "adjacency structure" of graphs

dodiscover/metrics.py
```diff
@@ -0,0 +1,66 @@
...
+    true_adj_mat = true_adj_mat > 0
+    pred_adj_mat = pred_adj_mat > 0
+
+    # vectorize and binarize for sklearn's confusion matrix
+    y_true = LabelBinarizer().fit_transform(true_adj_mat.flatten()).squeeze()
+    y_pred = LabelBinarizer().fit_transform(pred_adj_mat.flatten()).squeeze()
+
+    # compute the confusion matrix
+    conf_mat = confusion_matrix(y_true, y_pred, labels=[0, 1])
+    return conf_mat
```
robertness, line 68: What is the difference between what you're doing and a basic approach like the following?

```
G_true = nx.gnp_random_graph(5, .8)
G_detect = nx.gnp_random_graph(5, .8)
true_edges = set(G_true.edges)
detected_edges = set(G_detect.edges)

all_pairs = set([
    (a, b) for idx, a
    in enumerate(list(G_true.nodes))
    for b in list(G_true.nodes)[idx + 1:]
])

false_edges = all_pairs - true_edges

TP = true_edges.intersection(detected_edges)
FP = detected_edges - true_edges
TN = false_edges.intersection(detected_edges)
FN = true_edges - detected_edges

confusion_matrix = [
    [len(TP), len(FN)],
    [len(FP), len(TN)]
]
```

robertness (changes requested):
See comments.  Also, should there be a unit test?

dodiscover/metrics.py
robertness: I would say either you have an option for a strict matching of edge-type or just presence of an edge.  So in the latter case if the true edge is A <- B and the detected edge is A -> B or A - B, then this would be a false negative in the former case and a true positive in the latter.  I think you're just doing the latter case.  I think that is the preferred case and it's fine if we just have that.

dodiscover/metrics.py
adam2392: Nothing, except ordering. I can swap to this if you want?

I converted to a numpy array to leverage the `confusion_matrix` API in scikit-learn.

adam2392:
> See comments. Also, should there be a unit test?

Good catch. I added a unit-test, which caught some subtleties I think should be added.

dodiscover/metrics.py
adam2392: I think this can be easily extendable in the future with the inner code by adding an if/else statement for conversion to an undirected graph before the conversion to a numpy array.

robertness (approved):
LGTM.

PR #52, opened 2022-09-27
[ENH] Implementation of a "complete" FCI algorithm (augmented FCI)

adam2392:
Closes #32 

~Should be merged and rebased after #42 is merged.~

Changes proposed in this pull request:

- implements the FCI algorithm on top of the skeleton learning with latent confounders algorithm implemented in #50 
- This is a complete algorithm that implements R8-10, which is not present in causal-learn
- By changing the line within `LearnSkeleton` to `continue` instead of `break`, this fixes the issue with learning skeletons with PDS being "flaky" https://github.com/py-why/dodiscover/blob/c60cf2ef3a6b5dd05b0d5ad792ce27f1c150f695/dodiscover/constraint/skeleton.py#L388

Note: this is a relatively "large PR" simply because there is a lot of code that goes into implementing each of the rules of the FCI algorithm compared to e.g. the PC algorithm. Moreover, there is 500 LOC for unit tests. Lmk if you spot areas to trim.

TODO:
- [ ] It would be nice to demonstrate to the team in an integration test that R8-10 is able to orient more edges than causal-learn.

## Before submitting

- [ ] I've read and followed all steps in the [Making a pull request](https://github.com/py-why/dodiscover/blob/main/CONTRIBUTING.md#making-a-pull-request)
    section of the `CONTRIBUTING` docs.
- [ ] I've updated or added any relevant docstrings following the syntax described in the
    [Writing docstrings](https://github.com/py-why/dodiscover/blob/main/CONTRIBUTING.md#writing-docstrings) section of the `CONTRIBUTING` docs.
- [ ] If this PR fixes a bug, I've added a test that will fail without my fix.
- [ ] If this PR adds a new feature, I've added tests that sufficiently cover my new functionality.

## After submitting

- [ ] All GitHub Actions jobs for my pull request have passed.

PR #54, opened 2022-09-29
[ENH] Adding integration test with causal learn

adam2392:
Addresses TODO item in #52. 

Note in general it is somewhat challenging to compare algorithms wrt causal-learn because they do not instantiate the ability to pass in your own CI tests, nor do they allow you to use an "oracle". The best comparison would be done where we have an oracle for CI testing. I.e. the graph to query d-separation/m-separation.

- Adds an integration test wrt causal-learn for the FCI algorithm and PC algorithm

## Before submitting

## After submitting


adam2392:
This is good to merge now @robertness. 

If you want to take a look, the main changes are from the `tests/integration_tests/constraint/*.py` files, which show some integration tests against causal-learn's PC and FCI algorithm. 

The issue is that it seems our CI tests might be have implementation differences, especially the FisherZCITest. Because if we use the G^2 chi-square test, both repo's algorithms perform exactly the same. 

For the FCI algorithm, another issue will be that we have implemented rules 8-10, whereas causal-learn does not. But I'm doing a rather simple simulation, so I don' think that will be an issue.

Issue #56, opened 2022-10-02
Implement the Psi-FCI algorithm


adam2392:
**Is your feature request related to a problem? Please describe.**
The Psi-FCI algorithm is an extension of the FCI algorithm that is complete for Psi-PAGs. The algorithm can learn a Psi-PAG given a combination of observational and soft-interventional datasets. This will be a good introduction of a more complex multi-dataset algorithm that we can start taking advantage of to flush out the Context/data API. Moreover, it will help us test the limits of the existing graph and `dowhy.gcm` API.

We should implement the algorithm as outlined in https://causalai.net/r67.pdf.

**Describe the solution you'd like**
The Psi-FCI should subclass the existing FCI algorithm. In terms of graphical structure, there are additional graph API to consider. The main part being encoding what are known as "F-nodes". Another issue to sort out is how to test for conditional distribution differences (i.e. is P_1(y|x) = P_2(y|x)?), which is a central problem within Psi-FCI and other multi-dataset structure learning algos.

- [ ] Implement the KCD test https://arxiv.org/abs/2102.08208. This will allow us to test `P_1(y|x) = P_2(y|x)`
- [ ] Implement the Psi-PAG graph in pywhy-graphs
- [ ] Implement the Psi-FCI algorithm in dodiscover; propose necessary modifications to the general `fit(data, context)` / `learn_graph(data, context)` API
- [ ] Implement an example demonstrating Psi-FCI usage with a simulated SCM with soft interventions

**Additional context**
cc: @robertness @darthtrevino

robertness:
https://github.com/py-why/dodiscover/issues/59

PR #57, opened 2022-10-03
[DRAFT] Conditional discrepancy tests - Kernel approach and Bregman divergence

adam2392:
First step for #56 

Implements two new conditional discrepancy (CD) tests
- Kernel CD test: basically KCI but for testing `p_1(y|x) = p_2(y|x)`
- Bregman divergence CD test: uses divergences to test `p_1(y|x) = p_2(y|x)`

Still need to add unit tests. Once I get CI and tests working, I'll split this up into two PRs.

## Before submitting

## After submitting


dodiscover/cd/bregman.py
```diff
@@ -0,0 +1,142 @@
...
+    ):
+        rng = np.random.default_rng(random_state)
+
+        p = 0.5
+        n_samps = X.shape[0]
+
+        # compute the test statistic on the conditionally permuted
+        # dataset, where each group label is resampled for each sample
+        # according to its propensity score
+        null_dist = Parallel(n_jobs=self.n_jobs)(
```
darthtrevino, line 136: What parallelism mechanisms do we have currently in the library? Can users configure/tune this in any way?

dodiscover/cd/kernel_test.py
```diff
@@ -0,0 +1,249 @@
...
+
+        # TODO: CHECK THAT THIS IS CORRECT
+        K0 = K[np.ix_(first_mask, first_mask)]
+        K1 = K[np.ix_(second_mask, second_mask)]
+
+        # compute the number of samples in each
+        n0 = int(np.sum(1 - z))
+        n1 = int(np.sum(z))
+
+        if isinstance(self.l2, (int, float)):
```
darthtrevino, line 225: It kind of seems like this logic should be packed in a property getter. `self.get_regs()`

dodiscover/ci/kernel_test.py
```diff
@@ -151,24 +151,31 @@ def test(
             # when running CI, \ddot{X} comprises of (X, Z)
             X = np.concatenate((X, Z), axis=1)
 
-            Kz, sigma_z = self._compute_kernel(
+            Kz, sigma_z = compute_kernel(
                 Z,
-                distance_metric="l2",
+                distance_metric="euclidean",
```
darthtrevino, line 156: Would an enum make more sense than a string?

adam2392:
> Some minor nits, I think we should look into using a set of domain enumerations instead of depending on hardcoded string values for metrics and distance metrics.

Btw, migrating this PR into two separate PRs: #81 and #82

dodiscover/cd/kernel_test.py
adam2392: Will modify in #81

dodiscover/cd/bregman.py
adam2392: Parallelism is mainly done via `joblib`, so we do not use anything beyond the basic functionality. 

There are definitely many things we can add, such as local caching (to allow a user to warm-start their causal discovery algo.), or mem-mapping if they have a gigantic dataset, but these are relatively low priority imo until we get the algos/API implemented(?).

I might be missing other functionalities modern joblib has

Issue #66, opened 2022-12-08
Add notebook tutorial for PC algorithm


robertness:
**Is your feature request related to a problem? Please describe.**
We need notebooks for our algorithms.

**Describe the solution you'd like**
A notebook describing the FCI algorithm on a public benchmark dataset

Issue #46, opened 2022-09-08
[ENH] Add required/excluded edge directions in the orientation phase of PC/FCI


adam2392:
## Summary
Required and excluded edges are not taken into account at the orientation phase (i.e. setting edge directions) of constraint-based structure learning algos.

There are a few challenges though to tackle:
1. how to warn users that the resulting object may not obey the definition of a CPDAG/PAG
2. whether or not to set these in the skeleton phase (so they are not tested with CI tests), or to set these at the orientation level after CI tests have been run

## Misc. reference
"Required edges and excluded edges are built into the `LearnSkeleton` class rn meaning they do not account for directionality. However, these can also be taken into account at the orientation phase... 

There currently is a tradeoff that theory has yet to answer:
- inside skeleton phase: you just initially set the adjacencies based on this prior info. 
- inside orientation phase: you could orient additional edges

^ the nuanced challenge here is that the resulting object is NOT a CPDAG/PAG, and thus would not be expected to work with the downstream ID/Estimation algorithms that assume an input CPDAG/PAG. I'm leaning towards passing in a warning str to the graph object that when printed out, warns users to not use them in ID/estimation if they set prior edges that might break the conditional-independence statements.

_Originally posted by @adam2392 in https://github.com/py-why/dodiscover/pull/30#discussion_r966104788_"

jaron-lee:
Would this be the right place to include the knowledge tiers functionality that TETRAD currently has? E.g. I group variables into X groups, where variables in group i can only point to groups i+1 or greater. In some sense you could implement tiers by excluding a whole lot of edges but maybe a nicer API could be exposed to do this instead. 

This would also require directional excluded edges, so I would vote for doing so in the orientation phase.

jaron-lee:
Also is there a technical reason that would prevent us implementing both kinds of prior knowledge?  

As a concrete example I've done some structure learning applied to cardiac data. When talking to physicians the kinds of knowledge I was able to solicit was grouping variables into tiers (e.g. pre, intra, and post operation), and also ruling out certain edge directions (due to medical domain knowledge). I don't think we ruled out adjacencies between any variables in that particular example, but in general I could see this happening if the domain knowledge was weaker.

adam2392:
@jaron-lee 

[...]

PR #67, opened 2022-12-08
PC tutorial using ASIA data

robertness:
Fixes [issue 66](https://github.com/py-why/dodiscover/issues/66).

Adds a tutorial for the PC algo on a public dataset.

## How to review this PR

Here is the learned CPDAG on the ASIA network. 

image

Here is the comparable result from `pc.stable` in the bnlearn package.

image

So our performance is similar to bnlearn's implementation.  But it doesn't reconstruct the graph very well.  Here is the ground truth network for reference:

image

So it is not doing as well on this data, and therefore the tutorial is not telling a compelling story.  I suspect we could improve the causal discovery and the narrative if we add some constraints.  Any suggestions?

## Before submitting

- [x] If this PR fixes a bug, I've added a test that will fail without my fix.

## After submitting

- [x] All GitHub Actions jobs for my pull request have passed.


adam2392:
Oh interesting, so our PC algo incorrectly orients L -> S <- B, whereas it should be the other way around. This means that the algorithm incorrectly found that $L \perp B$ and $L \not\perp B | S$.

Are the alpha levels and CI tests the same used in dodiscover and R's pcstable?

Re constraints:
- We can impose the constraint that S must cause L and B because smoking causes lung cancer and not the other way around. We should note that the constraint may make the resulting graph not a valid CPDAG.
- Another thing we could do is just note this as an imperfection of causal discovery algorithms when data is unfaithful. I think IIRC, the ASIA dataset is unfaithful. This is seen because it is very hard to detect the edge between E and X.
- Another thing we can do is also run the ConservativePC to demonstrate that the CPDAG there is more "robust"

WDYT?

robertness:
> Are the alpha levels and CI tests the same used in dodiscover and R's pcstable?

I used .05 in bnlearn's pcstable.  What's the default here?

robertness:
> We can impose the constraint that S must cause L and B because smoking causes lung cancer and not the other way around. We should note that the constraint may make the resulting graph not a valid CPDAG.

So I once [created an algorithm](https://www.rdocumentation.org/packages/bnlearn/versions/4.6.1/topics/ctsdag) that would modify the CPDAG to account for edges fixed by interventions, constraints, and graph priors.  Do you think we could use something like that here?

robertness:
> * Another thing we could do is just note this as an imperfection of causal discovery algorithms when data is unfaithful. I think IIRC, the ASIA dataset is unfaithful. This is seen because it is very hard to detect the edge between E and X.

How about the ALARM network?

robertness:
> * Another thing we can do is also run the ConservativePC to demonstrate that the CPDAG there is more "robust"

Can you elaborate? How would this change things?

adam2392:
> How about the ALARM network? I think that was the first successful use case of the PC algo.

+1

> I used .05 in bnlearn's pcstable. What's the default here?

Also 0.05. Hmm perhaps there is a bug in the implementation of our CI test and/or the PC algo itself. Are you able to check through the separating sets? Cuz the skeleton learned looks the same in both, so this is good. Therefore the error must be in what the separating sets are or the orientation phase of the PC algo itself.

> So I once [created an algorithm](https://www.rdocumentation.org/packages/bnlearn/versions/4.6.1/topics/ctsdag) that would modify the CPDAG to account for edges fixed by interventions, constraints, and graph priors. Do you think we could use something like that here?

What does this do?

> Can you elaborate? How would this change things?

If we add background knowledge, the returned graph is no longer necessarily a Markov equivalence class of the DAG. It is an esoteric point for now, but it's something to note I would say. 

For example, say you get the true CPDAG:

```
X - Y - Z
```
and $X \not\perp Z, X \perp Z | Y$, then you apply prior knowledge to say $X \rightarrow Y$, then $X \rightarrow Y - Z$ is not a CPDAG for the CI statements written. 

In this simple setup, assuming the conditional independences learned were correct, then you would automatically have $X \rightarrow Y \rightarrow Z$, since a collider is not possible. But in general, I think this problem is open on how to systematically combine prior knowledge w/ causal orientation rules.

robertness:
> What does this do?

So for score-based algorithms, one learns a DAG and then converts it to a PDAG.  In bnlearn, you convert to a PDAG using [cpdag](https://github.com/cran/bnlearn/blob/4c0888088171d6394d377ede124d2ea6f4633117/R/frontend-bn.R#L121).  I notice now the algo has a wlbl argument which allows you to apply constraints that would force some edges to stay oriented.  This is what I was thinking of.  My algorithm extended from constraints to causal priors and interventions as well, but those extensions don't matter quite yet.
I suppose in a constraint algo like the PC algo, we would do something like have the constraints force certain edges to be oriented when otherwise they'd be undirected.  But we're already doing this via the contraints in the context object, correct?

robertness:
> We can impose the constraint that S must cause L and B because smoking causes lung cancer and not the other way around. We should note that the constraint may make the resulting graph not a valid CPDAG.

Ok, I tried imposing the constraint and it doesn't seem to work.  Is this a bug?

image

robertness:
Created an issue to address the need to convert characters to ints in the data: https://github.com/py-why/dodiscover/issues/69

robertness:
Created an issue to address the need to convert characters to ints in the data: https://github.com/py-why/dodiscover/issues/69

adam2392:
> > We can impose the constraint that S must cause L and B because smoking causes lung cancer and not the other way around. We should note that the constraint may make the resulting graph not a valid CPDAG.
> 
> Ok, I tried imposing the constraint and it doesn't seem to work. Is this a bug or did I make an error?

I did not thoroughly test adding constraints. However, I think this is a bug in the implementation, or perhaps just a miscommunication of how the constraints are applied. This is related to https://github.com/py-why/dodiscover/issues/46, which we should probably revisit.

I think the issue could be that inside the skeleton learning, we have the following function:

```
                    # ignore fixed edges
                    if (x_var, y_var) in self.context.included_edges.edges:
                        continue
```

robertness:
> I did not thoroughly test adding constraints. However, I think this is a bug in the implementation, or perhaps just a miscommunication of how the constraints are applied. This is related to #46, which we should probably revisit.
> 
> I think the issue could be that inside the skeleton learning, we have the following function:

I'm going to create a bug issue since we have reproducibility with my above code, and link to issue 46.

robertness:
@adam2392 I cleaned up the narrative to discuss the less than ideal results.  I created an issue to do another notebook that demonstrates the use of [constraints](https://github.com/py-why/dodiscover/issues/71), once that issue is fixed.  Can we merge?

adam2392:
> @adam2392 I cleaned up the narrative to discuss the less than ideal results. I created an issue to do another notebook that demonstrates the use of [constraints](https://github.com/py-why/dodiscover/issues/71), once that issue is fixed. Can you approve?

Hi @robertness I will try to get to this before EOY. So my hypothesis is that there is a runtime-issue that is created when you assume edge-constraints before the skeleton is discovered.

emrekiciman:
Approved.  Chatting with Robert, it looks like the notebook itself is working correctly, even though it is uncovering issues in the library.

emrekiciman (approved):
Approved.

adam2392 (changes requested):
@robertness I have a few comments that would just make the overall maintenance easier:

1. is it possible to use `bnlearn`'s API to pull the `asia.csv`, so that way it is not merged in with the source code?

```
# Example of interactive plotting
import bnlearn as bn

# Load example dataset
df = bn.import_example(data='asia')
```

Ref: https://erdogant.github.io/bnlearn/pages/html/Plot.html?highlight=asia

You would just delete `asia.csv` file and then add `bnlearn` to the doc dev list of dependencies: `[tool.poetry.group.docs.dependencies]` inside the `pyproject.toml` file.

2. I fixed the CI, so now there are some spelling issues caught in the notebook:

```
examples/notebooks/example-pc-algo.ipynb:155: distinquish ==> distinguish
[42](https://github.com/py-why/dodiscover/actions/runs/3728444238/jobs/6323464519#step:7:43)
examples/notebooks/example-pc-algo.ipynb:178: implemention ==> implementation
```

3. If you move the notebook to `doc/tutorial/markovian`, then this will cleanly separate simple example scripts and more involved Jupyter notebook tutorials. I would classify this notebook as a tutorial.

Lmk if you have any questions, or think something could be changed.

adam2392:
> Approved. Chatting with Robert, it looks like the notebook itself is working correctly, even though it is uncovering issues in the library.

Okay sounds good to me. I think the notebook itself is fine then. I left some minor comments to make sure the tutorials sections are relatively lightweight/clean. They should be easily resolved and then I'll approve and merge!

I'll work on debugging moving the edge constraints to post-skeleton-discovery.

robertness:
@adam2392 I moved to docs, removed the asia data, and added bnlearn as a dependency.

adam2392 (dismissed):

adam2392 (approved):
This now works. A few notes:

1. I fixed the CI. There were some installation issues that just needed to be iterated on based on the error messages that were provided.
2. I updated the notebook and in turn found a minor bug to fix in the PC algorithm orientation phase.
3. In the future, @robertness can you push to a branch on your fork and then start a PR from the fork rather than a branch on the main repo? 

Unfortunately, unsure why the Windows build is failing. It seems in general poetry has difficulty with Windows... @darthtrevino any ideas?

PR #85, opened 2023-01-09
[ENH] Classifier CMI test

adam2392:

Addresses the CMI part of #17 
Closes: #61 

- Implements a classifier based approach for estimating CMI, which in turn can be used for CI testing
- Refactors kernel and monte-carlo utility functions from its original `utils.py` location
- Refactors computing training/testing dataset for CI testing into a i) generate kNN function, ii) permute NN function, iii) partition dataset function. These can be used independently in different algorithms (e.g. CMIKNN, CCIT, CCMI).
- Reformatted GIN test files with `black`

Note: in the publication, they also support using a neural network approach, but that is outside the scope of this PR.

## Before submitting

## After submitting


dodiscover/ci/ccmi_test.py
```diff
@@ -0,0 +1,107 @@
...
+        # compute the final CMI value
+        val = (
+            scipy.special.digamma(knn_here)
+            - (
+                scipy.special.digamma(k_xz)
+                + scipy.special.digamma(k_yz)
+                - scipy.special.digamma(k_z)
+            ).mean()
+        )
+        return val
```
robertness, line 107: add a newline

robertness (approved):
The binary p-value is weird.  The reference paper says that the CMI estimate itself is a substitute for the p-value.  Perhaps return the CMI estimate?  It also says you could do a type of bootstrap estimate of the p-value by applying this method to many permuted datasets.  Maybe explain this in the docstrings?

Other that that LGTM.

adam2392:
Now, I just need to add a unit-test

robertness (approved):
LGTM.  One comment.  The CMI value that is calculated as follows: 

`val = hxyz - (hxz + hyz - hz).mean()`

Doesn't this have an asymptotic chi-squared distribution under the null hypothesis? If so, should there be an option to calculate the p-value that way?

PR #97, opened 2023-01-13
[Example Draft] Add GES wrapper

### LINKED ISSUE #29: Implement GES using causal-learn

adam2392:

Fixes #29 

-

## Before submitting

## After submitting

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -1,2 +1,3 @@
 raison
 wee
+ges
--- a/dodiscover/score/__init__.py
+++ b/dodiscover/score/__init__.py
@@ -1,0 +1,1 @@
+from .ges_alg import GES
--- a/dodiscover/score/ges_alg.py
+++ b/dodiscover/score/ges_alg.py
@@ -1,0 +1,43 @@
+from typing import Callable, Optional, Union
+
+import networkx as nx
+import pandas as pd
+from pywhy_graphs.array.export import clearn_arr_to_graph
+
+from dodiscover.context import Context
+
+
+# XXX: see https://github.com/juangamella/ges
+class GES:
+    graph_: Optional[nx.DiGraph]
+
+    def __init__(
+        self,
+        scoring_method: Union[Callable, str] = "bic",
+        max_indegree: int = None,
+        **scoring_method_kwargs,
+    ) -> None:
+        self.scoring_method = scoring_method
+        self.max_indegree = max_indegree
+        self.scoring_method_kwargs = scoring_method_kwargs
+
+        self.graph_ = None
+
+    def fit(self, df: pd.DataFrame, ctx: Context):
+        from causallearn.search.ScoreBased.GES import ges
+
+        X = df.to_numpy()
+
+        # run causal-learn
+        ges_record = ges(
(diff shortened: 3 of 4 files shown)


adam2392:
@robertness no time to address this, but this is a prelim draft of GES and general score-based algo API. 

However, what we need is a unit-test. I'm thinking a newcomer contributor can port the unit-test from causal-learn perhaps and rewrite it with a CPDAG graph from `pywhy-graphs`.

jaron-lee:
Doesn't look like `causallearn.search.ScoreBased.GES.ges` has the ability to evaluate under an oracle scoring method. Ideally I am trying to match the quality of testing that we had in `fcialg.py`.

adam2392:
> Doesn't look like `causallearn.search.ScoreBased.GES.ges` has the ability to evaluate under an oracle scoring method. Ideally I am trying to match the quality of testing that we had in `fcialg.py`.

In this case, we would need to replicate numerical tests. Idr, but does causal-learn have this? We can perhaps take those and just refashion the intput/output as the dodiscover api.

If not, okay maybe might be somewhat difficult to add this rn w/o assurances that the implementation is correct.

jaron-lee:
> > Doesn't look like `causallearn.search.ScoreBased.GES.ges` has the ability to evaluate under an oracle scoring method. Ideally I am trying to match the quality of testing that we had in `fcialg.py`.
> 
> In this case, we would need to replicate numerical tests. Idr, but does causal-learn have this? We can perhaps take those and just refashion the intput/output as the dodiscover api.
> 
> If not, okay maybe might be somewhat difficult to add this rn w/o assurances that the implementation is correct.

Hesitant to add numerical tests, causal-learn has these but in the test files there are comments where they note the test and implementation can be correct but the test fails because of essentially a random seed (and to open a PR if that happens). Sounds like a recipe for fragile tests, I much prefer the strategy of only testing with an oracle in the search algorithms, and leave the correctness of the constraint/scoring algorithm to that respective unit test.

jaron-lee:
The heavyweight option would be to try and build a similar abstraction for scoring algorithms like you did for conditional independence tests.

A more lightweight approach would be to just write an oracle scoring function in the style of causal-learn's and include that somewhere for testing. 

Bigger picture is that all of this (and indeed of efforts to integrate stuff from other libraries into dodiscover) depend on the conversation about the future of causal-learn/dodiscover.

adam2392:
Is it possible to use an Oracle in a Score-based algorithm? I'm not familiar w/ the score based procedures as much.

jaron-lee:
I'm speaking abstractly but I say oracle in the sense of a scoring algorithm that doesn't depend on a particular dataset but on the true model or at least of the true data generating distribution (maybe you would need an actual distribution so you can evaluate a true likelihood. 

Let me think about this more however.

PR #106, opened 2023-02-02
[WIP] Implement R5-7 for FCI

### LINKED ISSUE #53: Implement R5-7 for selection bias in FCI algorithm
**Is your feature request related to a problem? Please describe.**
Currently, the FCI algorithm implementation does not implement R5-8 in the FCI algorithm, which accounts for learning graphs with selection bias.

It would be a straightforward issue for someone to add these into the existing implementation. 

**Describe the solution you'd like**
Implement R5-8 of the FCI algorithm. See [publication](https://reader.elsevier.com/reader/sd/pii/S0004370208001008?token=5E04A22E2DE4FB1F3406C7CA0456B121D2E8E0019892D2EDB3260BB614D415B4DF520F9E79670883E8F2586A90E6143C&originRegion=us-east-1&originCreation=20220928184705).

- Implement `_apply_rule5`, `_apply_rule6` and `_apply_rule7`
- Update logic inside `_apply_rules_1to10`
- add unit tests to `tests/unit_tests/constraint/test_fcialg.py`
- [Bonus] Implement an example going through the different caveats and concepts of the FCI algorithm

Probably the most involved would be implementing the unit tests. The current graph API is a work in progress.

jaron-lee:
Fixes #53

- Implement R5-7 to handle selection bias
- Implement tests for R5-7 cases
- Change behavior of R8 to allow for selection bias possibility
- Modify `_apply_rules_1to10` to account for new rules
- Fix minor bug in R9 detected due to test case from Fig.6 Zhang 2008
- Change default behavior of algorithm to assume selection bias

## Before submitting

- [x] If this PR fixes a bug, I've added a test that will fail without my fix.

## After submitting


jaron-lee:
The PR is not quite complete but I had some questions regarding implementation:

* I want to write a test for a graph that has selection bias. Fig 1. of Zhang 2008 seems like a good candidate. Is there a way to represent selection in the `ADMG` class, so that the `Oracle` function behaves as we would like it to?
image
* The default setting for the parameter `selection_bias` in `FCI()` is False, but we could could change this to default True (aligns with causal-learn)

adam2392:
> * I want to write a test for a graph that has selection bias. Fig 1. of Zhang 2008 seems like a good candidate. Is there a way to represent selection in the `ADMG` class, so that the `Oracle` function behaves as we would like it to?

For public ref: discussion in https://github.com/py-why/pywhy-graphs/issues/43. 

> * The default setting for the parameter `selection_bias` in `FCI()` is False, but we could could change this to default True (aligns with causal-learn)

Yes we can def change the default to True as a result of this PR.

dodiscover/constraint/fcialg.py
```diff
@@ -398,14 +398,135 @@ def _apply_rule4(
...
         return added_arrows, explored_nodes
 
+    def _apply_rule5(self, graph: EquivalenceClass, u: Column, a: Column) -> bool:
+        """Apply rule 5 of FCI algorithm.
+
+        For each A o-o U, if there is an uncovered (e.g. every triple is unshielded)
+        circle path p = <A, B, ..., D, U> such that A, D are not adjacent and U, B are
+        not adjacent, then orient A o-o U and every edge on p as undirected.
+
+        Note that unlike the other rules, R5 is a binary operator.
```
adam2392, line 408: What does it mean that R5 is a binary operator?

jaron-lee: In the sense that all the other rules are defined on or otherwise change something in a triplet, but R5 is defined with respect to only two variables

adam2392:
Re CI failures: 

The unit tests might be pulling from dependency cache. You can prolly run: `poetry update pywhy-graphs` to update the poetry lock file to pull the latest version.

dodiscover/constraint/fcialg.py
adam2392, line 408:
```suggestion
        Note that unlike the other rules, R5 is a binary operator and operates on two nodes, rather than a triplet.
```

Just to be explicit.

tests/unit_tests/constraint/test_fcialg.py
```diff
@@ -559,3 +674,82 @@ def test_fci_complex(self):
...
+
+        expected_pag = PAG()
+        expected_pag.add_edge("A", "Ef", expected_pag.circle_edge_name)
+        expected_pag.add_edge("Ef", "A", expected_pag.circle_edge_name)
+        expected_pag.add_edge("Ef", "R", expected_pag.circle_edge_name)
+        expected_pag.add_edge("R", "Ef", expected_pag.circle_edge_name)
+        expected_pag.add_edge("R", "D", expected_pag.circle_edge_name)
+        expected_pag.add_edge("D", "R", expected_pag.circle_edge_name)
+        assert pag.nodes() == expected_pag.nodes()
+        assert pag.edges() == expected_pag.edges()
```
adam2392, line 753: So the expected PAG only has circle edges? Is there an example of a graph, where we know the outcome PAG has "tails" that we can test against?

adam2392 (changes requested):
You can add an abstract method with the signature `def add_edge(self, x, y, **kwargs)` to `dodiscover/_protocol.py:EquivalenceClass`

tests/unit_tests/constraint/test_fcialg.py
jaron-lee: I had another example that I ended up moving into the test for rule 9, but lemme see if i can think of some more complicated examples. If you have any nice examples please let me know.

I will try and implement:
- Fig 2. of https://arxiv.org/pdf/1104.5617.pdf

adam2392:
> Actually is there a reason why we would add this to `EquivalenceClass` as opposed to the parent `Graph`? That class seems to contain other functions for manipulating edges (e.g. `remove_edge`).

Ah yeah good point. Perhaps lets add to `Graph`.

tests/unit_tests/constraint/test_fcialg.py
adam2392: Fig 2 in that paper doesn't have selection bias I think, but tbh I have never seen a PAG example w/ selection bias lol. If it's too hard to think of an example, or find one, then I am okay w/o it. Or we can just take the graph for that PAG in Fig 2 and arbitrarily add a relevant set of nodes that will induce R5, and/or 6(?)

jaron-lee: Yeah maybe I will just do that for Fig. 2. 

We should probably address this in dodiscover documentation! How are people supposed to use this stuff if there are no good examples anywhere on the internet??

adam2392: Most definitely! I think our repo is starting to come together, where the next steps will be coming up w/ real (albeit simulations most likely) graphical models that reflect how FCI works.

Might be a good mental exercise to "work backwards a valid PAG to the DAG" lolol.

adam2392:
Will auto-merge if the CIs pass.

adam2392 (approved):
LGTM. thanks @jaron-lee!

PR #111, opened 2023-02-08
[ENH] Psi-FCI algorithm for handling interventional data

adam2392:
Signed-off-by: Adam Li <adam2392@gmail.com>

Fixes #56 
Closes: https://github.com/py-why/dodiscover/issues/62

This is the graph that is learned by Psi-FCI:

psi_pag

This is a relatively "larger" PR because it requires adding new skeleton, new discovery and also an example. I can break up this PR, but the main feedback I'm looking for is high-level at this point. The main portions are the example and the API, which are reviewable in the LOC in the unit-tests and example file.

Changes proposed in this pull request:
- Adds the Psi-FCI (or I-FCI) algorithm that leverages the PsiPAG and IPAG graph class in `pywhy-graphs`
- Improves the API for `Context` and `ContextBuilder` and tightens up their expected usage with unit tests
- Adds an example/tutorial showing how PsiFCI applied to Sachs dataset: https://output.circle-artifacts.com/output/job/dd40f7bf-1581-4cc4-a480-67f6e0fe9c41/artifacts/0/dev/auto_examples/plot_psifci_alg.html#sphx-glr-auto-examples-plot-psifci-alg-py

- ~Drafts a preliminary user guide for constraint algos and CI testing~

- [x] Compare against skeleton of the graphs in Figure 2 of https://causalai.net/r67.pdf
- [x] Compare against the learned causal graph in Figure 2 of https://causalai.net/r67.pdf
- [x] Compare performance in practice with the Sachs figure in supplemental of https://causalai.net/r67.pdf. Will need to pull in `bnlearn` and the preprocessed Sachs data to fully reproduce that.
- [x] Add example possibly showing computation on Sachs dataset?

## Before submitting

- [ ] I've read and followed all steps in the [Making a pull request](https://github.com/py-why/dodiscover/blob/main/CONTRIBUTING.md#making-a-pull-request)
    section of the `CONTRIBUTING` docs.
- [ ] I've updated or added any relevant docstrings following the syntax described in the
    [Writing docstrings](https://github.com/py-why/dodiscover/blob/main/CONTRIBUTING.md#writing-docstrings) section of the `CONTRIBUTING` docs.
- [ ] If this PR fixes a bug, I've added a test that will fail without my fix.
- [ ] If this PR adds a new feature, I've added tests that sufficiently cover my new functionality.

## After submitting

- [ ] All GitHub Actions jobs for my pull request have passed.

Issue #22, opened 2022-08-22
Documentation for new users - tutorials and examples


adam2392:
Reposting from @bloebp and Jaron

https://github.com/cmu-phil/example-causal-datasets for example datasets

The below for real/synthetic datasets used in dowhy.

"You can generate new ones (see the code at the bottom). Otherwise, the notebook uses this:
Latencies:
https://github.com/py-why/dowhy/blob/main/docs/source/example_notebooks/rca_microservice_architecture_latencies.csv

Supply chain:
https://github.com/py-why/dowhy/blob/main/docs/source/example_notebooks/supply_chain_week_over_week.csv"

### Proposal

Based on the larger discussion today, we like toy examples into the `examples/` directory for communicating core concepts and then `tutorials/` for larger more involved code and documentation that works with more complex datasets, such as the above.

adam2392:
Adding some more links: 

- https://py-why.github.io/dowhy/v0.8/example_notebooks/gcm_supply_chain_dist_change.html
- https://py-why.github.io/dowhy/v0.8/example_notebooks/rca_microservice_architecture.html

PR #126, opened 2023-03-17
[DOC] Adding user-guide docs for CI testing and constraint-based discovery

adam2392:
Breaking up the PR of #111. This is PR number 2.

Towards #22 

If/when we port to causal-learn, this is easily copyable. This relies on some upstream changes that need to occur in #127. Then the CI will get fixed.

- Adds some user-guide for CI testing and others

## Before submitting

## After submitting

### DIFF
--- a/doc/conditional_independence.rst
+++ b/doc/conditional_independence.rst
@@ -1,32 +1,40 @@
 .. _conditional_independence:
 
-========================
-Conditional Independence
-========================
+============
+Independence
+============
 
 .. currentmodule:: dodiscover.ci
 
-Testing for invariances in the data, such as conditional independence (CI) can be represented graphically
-in the form of d-separation statements under the causal faithfulness assumption. Given this, one
-is interested in high-powered and well-controlled CI tests that can be used to test for CI in data.
+Probabilistic independence among two random variables is when the realization of one
+variable does not affect the distribution of the other variable. It is a fundamental notion
+in probability and statistics that is used to determine if information about some variables
+can be gleaned from observations of other variables. Independence can be tested statistically
+in the form of unconditional (or marginal) independence, or conditional independence (CI).
+In the following, we present a brief overview of the various approaches to CI testing, where
+the marginal case is a special case when the conditioning set is empty.
 
-CI tests are framed as a statistical hypothesis test, with the following null hypothesis for a given
-pair of variables ``(X, Y)`` and a conditioning set ``Z`` (which may be empty):
+Conditional independence (CI) tests are framed as a statistical hypothesis test, with the following null hypothesis for a given
+pair of variables ``(X, Y)`` and a conditioning set ``Z`` (which may be empty). The null hypothesis is that X and Y are statistically
+independent given a conditioning set Z, i.e.,
 
 :math:`H_0: X \perp Y | Z`, or written in terms of their distribution :math:`H_0: P(Y | X, Z) = P(Y | Z)`
 
 Similarly, the alternative hypothesis is written as:
 
-:math:`H_0: X \not\perp Y | Z`, or written in terms of their distribution :math:`H_0: P(Y | X, Z) \neq P(Y | Z)`
+:math:`H_A: X \not\perp Y | Z`, or written in terms of their distribution :math:`H_A: P(Y | X, Z) \neq P(Y | Z)`
 
 Then typically, one posits an acceptable upper bound on the Type I error rate (false positive), typically :math:`\alpha = 0.05`
 and then either attempts to sample from the null distribution, or characterizes the asymptotic distribution
 of the test statistic. In both approaches a pvalue is computed, which is compared to :math:`\alpha`. The pvalue
 states the "probability of observing a test-statistic at least as extreme as our observed test-statistic in null distribution". By rejecting
 the null hypothesis, one claims that :math:`X \not\perp Y | Z`, so that X and Y are in fact (conditionally)
-dependent given Z. Note that if one fails to reject the null hypothesis, it is simply by convention that we
-claim X and Y are conditionally independent. It is not necessarily the case, and it is plausible that there
-is a weak dependency that is unable to be captured by our proposed CI test, and/or data samples.
+dependent given Z.
+
+Note that if one fails to reject the null hypothesis, we cannot accept the alternative hypothesis of
+independence strictly speaking. However, in practice in many settings, such as in causal discovery,
+we would still conclude that they are independent. It is not necessarily the case though, and
+it is plausible that there is a weak dependency that is unable to be captured by our proposed CI test, and/or data samples.
 
 It is because of this reason, one would typically like the most powerful test given assumptions about
 the data. With that in mind, there are various approaches to CI testing that are typically more powerful
@@ -45,7 +53,8 @@ on the underlying distributions. Unfortunately, CMI is notoriously difficult to 
 various proposals in the literature for estimating CMI, which we summarize here:
 
 - The Kraskov, Stogbauer and Grassberger (KSG) estimate approach estimates mutual information
-  via k-NN statistics :footcite:`kraskov_estimating_2004`. It was generalized to CMI by :footcite:`frenzel_partial_2007`.
+  via nearest-neighbor statistics :footcite:`kraskov_estimating_2004`. It computes nearest-neighbors
+  using a kNN algorithm. It was generalized to CMI by :footcite:`frenzel_partial_2007`.
   This class of estimators is asymptotically correct, meaning if we had an infinite amount of data
   we would obtain the true value of the CMI. However, it relies on statistics generated from the k-NN,
   which if we implement the naive approach using a KDTree, then it generally suffers in high-dimensions.
@@ -56,10 +65,10 @@ various proposals in the literature for estimating CMI, which we summarize here:
   and then the CMI value generated from the actual observed data can be compared to the CMI values computed
   on the permutated datasets to estimate a pvalue.
 
-  It is worth noting that if one has "robust" estimates of k-NN using for example a model that
-  is effective in high-dimensions, then the KSG estimator for CMI may be quite good. For example,
+  It is worth noting that if one has good estimates of nearest-neighbors using for example a model that
+  is effective in high-dimensions, then the KSG estimator for CMI may still be effective. For example,
   one can use variants of Random Forests to generate adaptive nearest-neighbor estimates in high-dimensions
-  or on manifolds, such that the KSG estimator is still quite good.
+  or on manifolds, such that the KSG estimator is still powerful.
 
 .. autosummary::
    :toctree: generated/
@@ -81,7 +90,7 @@ Partial (Pearson) Correlation
 -----------------------------
 Partial correlation based on the Pearson correlation is equivalent to CMI in the setting
 of normally distributed data. Computing partial correlation is fast and efficient and
-thus attractive to use. However, this **relies on the assumption of multivariate Gaussianity**,
+thus attractive to use. However, this **relies on the assumption that the variables are Gaussiany**,
 which may be unrealistic in certain datasets.
 
 .. autosummary::
@@ -117,11 +126,19 @@ estimate a pvalue.
 Classifier-based Approaches
 ---------------------------
 Another suite of approaches that rely on permutation testing is the classifier-based approach.
-By shuffling data in an intelligent way, one can setup a hypothesis test for CI based on the
+By shuffling the data, one can setup a hypothesis test for CI based on the
 predicted probabilities from a classification-model. Intuitively, if the shuffled data is similar 
 to the unshuffled data, such that the classification-model achieves non-trivial performance
 (e.g. >50\% accuracy on a balanced dataset), then one fails to reject the null hypothesis and would
 state that the original data was in fact CI :footcite:`Sen2017model`.
+
+When performing marginal independence testing between two sets of variables, ``X`` and ``Y``,
+it is sufficient to shuffle data by just permuting either ``X`` rows or ``Y`` rows uniformly.
+When performing CI testing conditioned on a third set of variables ``Z``, one must perform what is known
+as conditional shuffling. One computes the nearest-neighbors in the ``Z`` subspace and then
+permutes rows based on samples that are close in ``Z`` subspace, which
+helps maintain dependence between (X, Z) and (Y, Z) (if it exists), but generates a
+conditionally independent dataset.
 
 .. autosummary::
    :toctree: generated/
--- a/doc/conditional_independence.rst
+++ b/doc/conditional_independence.rst
@@ -1,0 +1,180 @@
+.. _conditional_independence:
+
+========================
+Conditional Independence
+========================
+
+.. currentmodule:: dodiscover.ci
+
+Testing for invariances in the data, such as conditional independence (CI) can be represented graphically
+in the form of d-separation statements under the causal faithfulness assumption. Given this, one
+is interested in high-powered and well-controlled CI tests that can be used to test for CI in data.
+
+CI tests are framed as a statistical hypothesis test, with the following null hypothesis for a given
+pair of variables ``(X, Y)`` and a conditioning set ``Z`` (which may be empty):
+
+:math:`H_0: X \perp Y | Z`, or written in terms of their distribution :math:`H_0: P(Y | X, Z) = P(Y | Z)`
+
+Similarly, the alternative hypothesis is written as:
+
+:math:`H_0: X \not\perp Y | Z`, or written in terms of their distribution :math:`H_0: P(Y | X, Z) \neq P(Y | Z)`
+
+Then typically, one posits an acceptable Type I error rate (false positive), typically :math:`\alpha = 0.05`
+and then either attempts to sample from the null distribution, or characterizes the asymptotic distribution
+of the test statistic. In both approaches a pvalue is computed, which is compared to :math:`\alpha`. The pvalue
+states the "probability that we observe our data (e.g. test statistic) under the null hypothesis". By rejecting
+the null hypothesis, one claims that :math:`X \not\perp Y | Z`, so that X and Y are in fact (conditionally)
+dependent given Z. Note that if one fails to reject the null hypothesis, it is simply by convention that we
+claim X and Y are conditionally independent. It is not necessarily the case, and it is plausible that there
+is a weak dependency that is unable to be captured by our proposed CI test, and/or data samples.
+
+It is because of this reason, one would typically like the most powerful test given assumptions about
+the data. With that in mind, there are various approaches to CI testing that are typically more powerful
+with certain assumptions on the underlying data distribution.
+
+Conditional Mutual Information
+------------------------------
+Conditional mutual information (CMI) is a general formulation of CI, where CMI is defined as
+:math::
+    
+    \\int log \frac{p(x, y | z)}{p(x | z) p(y | z)}
+
+As we can see, CMI is equal to 0, if and only if :math:`p(x, y | z) = p(x | z) p(y | z)`, which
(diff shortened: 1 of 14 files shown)


doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
...
+CI tests are framed as a statistical hypothesis test, with the following null hypothesis for a given
+pair of variables ``(X, Y)`` and a conditioning set ``Z`` (which may be empty):
+
+:math:`H_0: X \perp Y | Z`, or written in terms of their distribution :math:`H_0: P(Y | X, Z) = P(Y | Z)`
+
+Similarly, the alternative hypothesis is written as:
+
+:math:`H_0: X \not\perp Y | Z`, or written in terms of their distribution :math:`H_0: P(Y | X, Z) \neq P(Y | Z)`
+
+Then typically, one posits an acceptable Type I error rate (false positive), typically :math:`\alpha = 0.05`
```
robertness, line 22: nit: "acceptable _upper bound_ on the type I error rate"

doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
...
+and then either attempts to sample from the null distribution, or characterizes the asymptotic distribution
+of the test statistic. In both approaches a pvalue is computed, which is compared to :math:`\alpha`. The pvalue
+states the "probability that we observe our data (e.g. test statistic) under the null hypothesis". By rejecting
```
robertness, line 25: "probability of observing a t-statistic at _least as extreme_ as our observed t-statistic in null distribution"

doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
...
+
+- Direct posterior estimates can be implemented with a classification model by directly
+  estimating :math:`P(y|x)` and :math:`P(y|x,z)`, which can be used as plug-in estimates
+  to the equation for CMI.
+
+Partial (Pearson) Correlation
+-----------------------------
+Partial correlation based on the Pearson correlation is equivalent to CMI in the setting
+of normally distributed data. Computing partial correlation is fast and efficient and
+thus attractive to use. However, this **relies on the assumption of multivariate Gaussianity**,
```
robertness, line 84: "Gaussianity"?

doc/constraint_causal_discovery.rst
```diff
@@ -0,0 +1,132 @@
...
+:math:`H_0: X \perp Y | Z` and :math:`H_A: X \not\perp Y | Z`
+
+For a given node ``X`` and ``Y`` in the underlying causal graph of interest, and
+a conditioning set, ``Z``. CI tests can have a variety of different assumptions
+that make one better than another in different settings and data assumptions.
+For more information on the CI tests themselves, see :ref:`conditional_independence`.
+
+Fundamental Assumptions of Constraint-Based Causal Discovery
+------------------------------------------------------------
+The fundamental assumptions of all algorithms in this section are the Markov property assumption
```
robertness, line 26: an no latents?

robertness, line 26: I can see constraint-based working with latents. But if there are latents, I guess Markov wouldn't hold, unless using a mixed graph?

doc/constraint_causal_discovery.rst
```diff
@@ -0,0 +1,132 @@
...
+and the causal faithfulness assumption :footcite:`Spirtes1993` and :footcite:`Pearl_causality_2009`.
+
+The Markov assumption states that all d-separation statements in the causal graph imply a
+corresponding CI statement in the data. This is a core-assumption that users from graphical modeling
+may be familiar with.
+
+On the other hand, the causal faithfulness assumption states that all
```
robertness, line 33: I think I read somewhere that the chances of a faithfulness violation has measure zero when there are no latents and no deterministic relationships between variables.  Do we want to say that faithfulness violations are not that big a deal relative to violations of other assumptions

doc/constraint_causal_discovery.rst
```diff
@@ -0,0 +1,132 @@
...
+CI statements in the data map to a d-separation statement. That is, there are no accidental
+CI that occur in the data, which are not represented by a d-separation statement in the underlying
+causal graph. The causal faithfulness assumption is a very problematic assumption because in practice
+one might have data that is very weakly dependent, such that a CI test under a specified :math:`\alpha`
+level would fail to reject the null hypothesis and conclude the variables in question are CI. In higher
```
robertness, line 38: Meh.  So you get a sparser graph.  I'd say a denser graph is more probabilistic in causal inference.  You get edges where there is no causality.

doc/references.bib
```diff
@@ -53,6 +64,144 @@ @book{Neapolitan2003
...
...on binnings, they are based on entropy estimates from k-nearest neighbor distances. This means that they are data efficient (with k=1 we resolve structures down to the smallest possible scales), adaptive (the resolution is higher where data are more numerous), and have minimal bias. Indeed, the bias of the underlying entropy estimates is mainly due to nonuniformity of the density at the smallest resolved scale, giving typically systematic errors which scale as functions of k∕N for N points. Numerically, we find that both families become exact for independent distributions, i.e. the estimator ˆM(X,Y) vanishes (up to statistical fluctuations) if μ(x,y)=μ(x)μ(y). This holds for all tested marginal distributions and for all dimensions of x and y. In addition, we give estimators for redundancies between more than two random variables. We compare our algorithms in detail with existing algorithms. Finally, we demonstrate the usefulness of our estimators for assessing the actual independence of components obtained from independent component analysis (ICA), for improving ICA, and for estimating the reliability of blind source separation., This article appears in the following collections:},
```
robertness, line 178: Maybe remove abstract

robertness (dismissed):
LGTM

doc/conditional_independence.rst
adam2392, line 22:
```suggestion
Then typically, one posits an acceptable upper bound on the Type I error rate (false positive), typically :math:`\alpha = 0.05`
```

doc/conditional_independence.rst
adam2392, line 25:
```suggestion
states the "probability of observing a test-statistic at least as extreme as our observed test-statistic in null distribution". By rejecting
```

doc/conditional_independence.rst
adam2392: Seems like it is a word? https://en.wikipedia.org/wiki/Non-Gaussianity hehe. Any suggestions?

doc/constraint_causal_discovery.rst
adam2392: The Markov assumption states that d-separation statements imply conditional independences. These still hold even in the presence of bidirected edges?

robertness (dismissed):

doc/constraint_causal_discovery.rst
adam2392: Sure that is fair, but for dealing with real data, the problem of "weak-faithfulness" is highly likely. I can add some discussion around that.

ref: https://www.jstor.org/stable/23566569?casa_token=kzePz4xv42MAAAAA:PuhX4bXVGqes-QdqLTysC1wTY_gie2IwDRda1TACReQsu4g8mh3ewb_Hw_4mu7mzP3L8G_PW-Cgf7SMuJYL2UIwANUcVu3xVgGweU6_k0X9GjtX4LgE

doc/constraint_causal_discovery.rst
adam2392: Are you saying that having no removal of edges is more problematic?

doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
...
+
+========================
+Conditional Independence
+========================
+
+.. currentmodule:: dodiscover.ci
+
+Testing for invariances in the data, such as conditional independence (CI) can be represented graphically
+in the form of d-separation statements under the causal faithfulness assumption. Given this, one
+is interested in high-powered and well-controlled CI tests that can be used to test for CI in data.
```
bloebp, line 11: While it is a good fit for dodiscover, I wouldn't make any causal or general graph reference at this point, since (conditional) independence tests are just hypothesis tests as mentioned in the next paragraph. However, it fits to the dodiscover story of course, but if we move it to the stats package, this part should probably be re-written/removed.

Maybe you can move this graphical remark to the constrained based causal discovery part?

doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
+.. _conditional_independence:
+
+========================
+Conditional Independence
```
bloebp, line 4: I am wondering if simply "Independence" would be better. It currently sounds like that the unconditional case is a special case of the conditional independence simply with the conditioning set being empty. Although you can definitely see it like that, I would probably still distinguish between them more clearly. Why not just introduce independence tests in general and mention that one can either perform a unconditional or conditional test?

doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
...
+.. currentmodule:: dodiscover.ci
+
+Testing for invariances in the data, such as conditional independence (CI) can be represented graphically
+in the form of d-separation statements under the causal faithfulness assumption. Given this, one
+is interested in high-powered and well-controlled CI tests that can be used to test for CI in data.
+
+CI tests are framed as a statistical hypothesis test, with the following null hypothesis for a given
+pair of variables ``(X, Y)`` and a conditioning set ``Z`` (which may be empty):
+
+:math:`H_0: X \perp Y | Z`, or written in terms of their distribution :math:`H_0: P(Y | X, Z) = P(Y | Z)`
```
bloebp, line 16: Write it out as a text as well: "The null hypothesis is that X and Y are statistically independent given a conditioning set Z, i.e., :math:`H_0: X \perp Y | Z`, or written in terms of their distribution :math:`H_0: P(Y | X, Z) = P(Y | Z)`"

doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
...
+
+Similarly, the alternative hypothesis is written as:
+
+:math:`H_0: X \not\perp Y | Z`, or written in terms of their distribution :math:`H_0: P(Y | X, Z) \neq P(Y | Z)`
```
bloebp, line 20: Shouldn't this be H_1 then? One should also emphasize that it is impossible to reject this in practice and, thus, we can only reject independence.

doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
...
+
+Then typically, one posits an acceptable upper bound on the Type I error rate (false positive), typically :math:`\alpha = 0.05`
+and then either attempts to sample from the null distribution, or characterizes the asymptotic distribution
+of the test statistic. In both approaches a pvalue is computed, which is compared to :math:`\alpha`. The pvalue
+states the "probability of observing a test-statistic at least as extreme as our observed test-statistic in null distribution". By rejecting
+the null hypothesis, one claims that :math:`X \not\perp Y | Z`, so that X and Y are in fact (conditionally)
+dependent given Z. Note that if one fails to reject the null hypothesis, it is simply by convention that we
+claim X and Y are conditionally independent. It is not necessarily the case, and it is plausible that there
```
bloebp, line 28: Maybe formulate the "... that we claim X and Y are conditionally independent."-part a bit weaker. Something like "Note that if one fails to reject the null hypothesis, we cannot accept the alternative hypothesis of independence strictly speaking. However, in practice, such as in causal discovery, we would still conclude that they are independent." or something along the line.

doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
...
+
+  Estimates of CMI can be converted into a CI hypothesis test by permutation testing :footcite:`Runge2018cmi`.
+  One can generate estimated samples from the null distribution by permuting samples in an intelligent manner
+  and then the CMI value generated from the actual observed data can be compared to the CMI values computed
+  on the permutated datasets to estimate a pvalue.
+
+  It is worth noting that if one has "robust" estimates of k-NN using for example a model that
+  is effective in high-dimensions, then the KSG estimator for CMI may be quite good. For example,
+  one can use variants of Random Forests to generate adaptive nearest-neighbor estimates in high-dimensions
+  or on manifolds, such that the KSG estimator is still quite good.
```
bloebp, line 62: I would remove the "quite good"s here. Just say that it can be 'powerful' even in high dimensions if the knn estimator can capture it.

doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
...
+  we would obtain the true value of the CMI. However, it relies on statistics generated from the k-NN,
+  which if we implement the naive approach using a KDTree, then it generally suffers in high-dimensions.
+  In our examples, we see it suffer with dimensionality > 4 or 5.
+
+  Estimates of CMI can be converted into a CI hypothesis test by permutation testing :footcite:`Runge2018cmi`.
+  One can generate estimated samples from the null distribution by permuting samples in an intelligent manner
+  and then the CMI value generated from the actual observed data can be compared to the CMI values computed
+  on the permutated datasets to estimate a pvalue.
+
+  It is worth noting that if one has "robust" estimates of k-NN using for example a model that
```
bloebp, line 59: What does "robust estimates of k-NN" mean? I guess you mean the quantity that you estimate using k-NN (e.g., mutual information), maybe just focus on that. Otherwise, k-NN would simply tell you the k neighbors and that just depends on your distance measure.

doc/conditional_independence.rst
bloebp: I think Gaussianity is quite common, but the combination with "multivariate" sounds odd. What about "However, this relies on the assumption that the variables are Gaussian."

doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
...
+By shuffling data in an intelligent way, one can setup a hypothesis test for CI based on the
+predicted probabilities from a classification-model. Intuitively, if the shuffled data is similar 
+to the unshuffled data, such that the classification-model achieves non-trivial performance
+(e.g. >50\% accuracy on a balanced dataset), then one fails to reject the null hypothesis and would
+state that the original data was in fact CI :footcite:`Sen2017model`.
+
+.. autosummary::
+   :toctree: generated/
+
+    ClassifierCITest
```
bloebp, line 129: Not relevant for this PR: But why isn't there one based on regression? So, imply checking whether the model improves when you add the other variable(s)?

doc/conditional_independence.rst
```diff
@@ -0,0 +1,180 @@
...
+
+.. autosummary::
+   :toctree: generated/
+
+    KernelCITest
+
+Classifier-based Approaches
+---------------------------
+Another suite of approaches that rely on permutation testing is the classifier-based approach.
+By shuffling data in an intelligent way, one can setup a hypothesis test for CI based on the
```
bloebp, line 120: What does this "intelligent" mean here? Wouldn't you just uniformly shuffle them in practice?

doc/constraint_causal_discovery.rst
```diff
@@ -0,0 +1,136 @@
...
+==================================
+.. automodule:: dodiscover.constraint
+    :no-members:
+    :no-inherited-members:
+.. currentmodule:: dodiscover
+
+The following are a set of methods intended for (non-parametric) structure learning
+of causal graphs (i.e. causal discovery) given observational and/or interventional data
+by checking constraints in the form of conditional independences (CI). At a high level
+all constraint-based causal discovery, tests CI statements, where we use different CI
```
bloebp, line 14: The "At a high level.." sentence is difficult to read. What about: "At a high level, all constraint-based causal discovery algorithms test CI statements, where we use..."

doc/constraint_causal_discovery.rst
```diff
@@ -0,0 +1,136 @@
...
+
+==================================
+Constraint-based causal discovery
+==================================
+.. automodule:: dodiscover.constraint
+    :no-members:
+    :no-inherited-members:
+.. currentmodule:: dodiscover
+
+The following are a set of methods intended for (non-parametric) structure learning
```
bloebp, line 11: Referencing to my previous comment in the CI part, here you could instead mention that CI correspond to d-separation in a graph etc.

doc/constraint_causal_discovery.rst
```diff
@@ -0,0 +1,136 @@
...
+that make one better than another in different settings and data assumptions.
+For more information on the CI tests themselves, see :ref:`conditional_independence`.
+
+Fundamental Assumptions of Constraint-Based Causal Discovery
+------------------------------------------------------------
+The fundamental assumptions of all algorithms in this section are the Markov property assumption
+and the causal faithfulness assumption :footcite:`Spirtes1993` and :footcite:`Pearl_causality_2009`.
+
+The Markov assumption states that all d-separation statements in the causal graph imply a
+corresponding CI statement in the data. This is a core-assumption that users from graphical modeling
```
bloebp, line 30: This is rather the faithfulness assumption and not the Markov assumption?

doc/constraint_causal_discovery.rst
bloebp: I agree with Adam, although faithfulness is very unlikely in practice, a weak form of it would already be an issue for the tests. And that can be quite common.

doc/conditional_independence.rst
adam2392: Done. Lmk wdyt.

doc/conditional_independence.rst
adam2392: Done. Lmk wdyt.

doc/conditional_independence.rst
adam2392: Yes good catch fixed.

doc/conditional_independence.rst
adam2392: Added.

doc/conditional_independence.rst
adam2392: Done.

doc/conditional_independence.rst
adam2392: Sorry I reworded it. So the KSG estimate of MI is based on kNN-based statistics. But it is a theoretical result that as dimensionality increases, the number of samples needed for kNN algorithm grows exponentially. Thus the MI estimate has at least the same if not worse behavior. Assuming you have an alternative algorithm for estimating nearest-neighbors that is robust to noise, then maybe the KSG estimate of MI performs better.

doc/conditional_independence.rst
adam2392: Modified

doc/conditional_independence.rst
adam2392: I just never implemented one because I saw dowhy had it. We can consolidate now that we moving CI stuff to pywhy-stats :). I think upon skimming, the meta-method for helping generate the pvalue can be consolidated with a suite of meta-methods for pvalue estimation.

doc/conditional_independence.rst
adam2392: Marginal ind testing = shuffling the data unifromly
CI testing = shuffling the data conditioned on Z (done via kNN). This is a subroutine in any permutation type procedure.

adam2392: I reworded this section tho

doc/constraint_causal_discovery.rst
adam2392: Done.

doc/constraint_causal_discovery.rst
adam2392: Isn't it

Markov assumption = d-separation -> CI
whereas faithfulness = CI -> d-separation?

adam2392: Oh I see the confusion. I will reword this section.

doc/constraint_causal_discovery.rst
adam2392: Done

adam2392:
@robertness this is good for review again

bloebp (approved):

PR #128, opened 2023-03-17
[ENH] Chisq set of categorical CI testing

### LINKED ISSUE #124: Add Chi-square CI tests and class of power-divergence tests
See scipy.stats and pgmpy.

https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.power_divergence.html#scipy.stats.power_divergence

adam2392:
Breaking up the PR of https://github.com/py-why/dodiscover/pull/111. This is PR number 3.

Fixes #124 

- Ports over suite of categorical CI tests from pgmpy.

There is some performance differences wrt the current G^2 implementation. Not sure why... Need to debug.

## Before submitting

## After submitting


adam2392:
This is also to then migrate over to dodiscover. I just had this PR open already so figured. Might as well get this in and then do the migration.

tests/unit_tests/conditional/ci/test_categorical_test.py
```diff
@@ -0,0 +1,156 @@
+import numpy as np
+import pandas as pd
+from numpy.testing import assert_almost_equal
+
+from dodiscover.ci import CategoricalCITest
+
+df_adult = pd.read_csv("dodiscover/testdata/adult.csv")
+
+
+def test_chisquare_adult_dataset():
```
bloebp, line 10: Optimally, these would be multiple tests with descriptive test names. For instance, the "given-when-then"-pattern is quite common for that: ```def given_unconditionally_independent_variables_when_perform_categorical_ci_test_with_pearson_then_returns_expected_test_statistic()```. This may look ridiculously long, but in case of unit tests, that is fine and probably less work then adding more descriptions to it.

Here, maybe you can at least make one test for the unconditional and one test for the conditional setting.

dodiscover/testdata/adult.csv
bloebp, line 1: Note: One should avoid putting csv files into the repository, but since we don't really have an external storage (or do we?), I guess its fine for now.

tests/unit_tests/conditional/ci/test_categorical_test.py
```diff
@@ -0,0 +1,156 @@
...
+        y_vars={"Income"},
+        z_covariates=["Age", "Education", "HoursPerWeek", "MaritalStatus"],
+        df=df_adult,
+    )
+    assert_almost_equal(coef, 65.59, decimal=1)
+    assert_almost_equal(p_value, 0.999, decimal=2)
+    assert ci_est.dof_ == 131
+
+
+def test_discrete_tests():
```
bloebp, line 74: Also here, the test name can be more descriptive and there should be a separate test per parameter. What about [parametrizing](https://docs.pytest.org/en/7.1.x/example/parametrize.html) the test instead?

tests/unit_tests/conditional/ci/test_categorical_test.py
```diff
@@ -0,0 +1,156 @@
...
+        #     t.test(
+        #         x_vars={"Education"},
+        #         y_vars={"MaritalStatus"},
+        #         z_covariates=["Age", "Sex"],
+        #         df=df_adult,
+        #     )[1] < 0.05
+        # )
+
+
+def test_exactly_same_vars():
```
bloebp, line 141: Same as in the previous tests.

dodiscover/ci/categorical_test.py
```diff
@@ -0,0 +1,200 @@
...
+                        f"Skipping the test {X} \u27C2 {Y} | {Z[0]}={z_state}. Not enough samples"
+                    )
+                else:
+                    z_str = ", ".join([f"{var}={state}" for var, state in zip(Z, z_state)])
+                    logging.info(f"Skipping the test {X} \u27C2 {Y} | {z_str}. Not enough samples")
+        p_value = 1 - stats.chi2.cdf(chi, df=dof)
+        import numpy as np
+
+        if np.isnan(p_value):
+            print(p_value, chi, c, dof, X, Y, Z)
```
bloebp, line 119: Avoid putting print statements. Maybe instead use a logger as in the lines above.

dodiscover/ci/categorical_test.py
```diff
@@ -0,0 +1,200 @@
...
+            c, _, d, _ = stats.chi2_contingency(
+                df.groupby([X, Y]).size().unstack(Y, fill_value=0), lambda_=lambda_
+            )
+            print(c, d)
```
bloebp, line 123: Same here

dodiscover/ci/categorical_test.py
```diff
@@ -0,0 +1,200 @@
...
+    True
+    >>> chi_square(X='A', Y='B', Z=['D', 'E'], data=data, boolean=True, significance_level=0.05)
+    False
+    """
+
+    # Step 1: Check if the arguments are valid and type conversions.
+    if hasattr(Z, "__iter__"):
+        Z = list(Z)
+    else:
+        raise (f"Z must be an iterable. Got object type: {type(Z)}")
```
bloebp, line 83: What type of exception? I guess ```raise ValueError(...)```?

Another remark: I don't think it necessary to test for these kind of things. Just add an appropriate typing to the variables and then assume that the correct data type is given. Otherwise, you will need to check all kind of things, which is not really worth it.

dodiscover/ci/categorical_test.py
```diff
@@ -0,0 +1,200 @@
...
+
+import pandas as pd
+from scipy import stats
+
+from dodiscover.ci.base import BaseConditionalIndependenceTest
+from dodiscover.typing import Column
+
+
+# copied from pgmpy
+def power_divergence(X, Y, Z, data, lambda_="cressie-read"):
```
bloebp, line 13: Types are missing.

Also, if you copied this from pgmpy, pay attention to their license (is it ok to copy and modify in the first place) and accordingly reference the license if required.

tests/unit_tests/conditional/ci/test_categorical_test.py
```diff
@@ -0,0 +1,156 @@
...
+            t.test(
+                x_vars={"Immigrant"},
+                y_vars={"Sex"},
+                z_covariates=[],
+                df=df_adult,
+            )[1]
+            < 0.05
+        )
+
+        # XXX: Test returns nan...
```
bloebp, line 122: Remove these commented lines. Also, if the test returns nan, make sure you understand why to avoid that there is actually a bug.

dodiscover/ci/categorical_test.py
jaron-lee: pgmpy is MIT licensed so it should be fine, I believe

bloebp: You still need to copy the license over accordingly:
https://github.com/pgmpy/pgmpy/blob/dev/LICENSE#L12

So, just copying the code without that would still not be ok.

adam2392: But all pywhy packages are MIT licensed, so I think this should be okay to just leave as is and provide credit assignment to pgmpy in the docstrings?

tests/unit_tests/conditional/ci/test_categorical_test.py
adam2392: Done.

dodiscover/testdata/adult.csv
adam2392: Yeah we can create a small test-data (i.e. files < 10MB ideally) repository in the future similar to https://github.com/mne-tools/mne-testing-data.

Then we can easily leverage `pooch` to just grab data during test time.

dodiscover/ci/categorical_test.py
bloebp: Strictly speaking, there is "Copyright (c) 2013-2021 pgmpy" in their MIT license, which we also strictly speaking need to copy 1:1.
Ok, at least emphasize that part and say it follows MIT license.

tests/unit_tests/conditional/ci/test_categorical_test.py
adam2392: Done

dodiscover/ci/categorical_test.py
adam2392: Okay that works too. Modified

dodiscover/testdata/adult.csv
adam2392: Perhaps something we can start discussing in the Monday meetings. It would be nice to consolidate a "small" testing repository that can ensure "correctness" of certain algorithms, while also being able to have a small Memory footprint and can run in very short period of time.

We can also use other packages like `bnlearn` for more "intensive" testing against datasets

bloebp (approved):

PR #129, opened 2023-03-27
Addition of topological order based methods

francescomontagna:
Implementation of four causal discovery methods on Additive Noise Models with observational data. The four methods follow the same _meta-algorithm_, first estimating the topological ordering of the nodes in the causal graph, then by pruning the fully connected adjacency matrix encoding of the inferred topological ordering. The four algorithms are

- CAM (Causal Additive Model) algorithm from [CAM: Causal additive models, high-dimensional order search and penalized regression](https://arxiv.org/abs/1310.1533), Buhlmann et al., 2013.
- SCORE algorithm from [Score Matching Enables Causal Discovery of Nonlinear Additive Noise Models](https://proceedings.mlr.press/v162/rolland22a.html), Rolland et al., 2022. 
- DAS (Discovery At Scale) algorithm from [Scalable Causal Discovery with Score Matching](https://openreview.net/forum?id=6VvoDjLBPQV), Montagna et al., 2023.
- NoGAM (Not only Gaussian Additive Model) algorithm from [Causal Discovery with Score Matching on Additive Models with Arbitrary Noise](https://openreview.net/forum?id=rVO0Bx90deu), Montagna et al., 2023.

All the addition concerning this pull request can be found in the `dodiscover/toporder` directory (plus the additional unit tests).

This contribution to the DoDiscover library is part of my internship work at Amazon AWS, Tubingen.

## Before submitting

- [X] I've read and followed all steps in the [Making a pull request](https://github.com/py-why/dodiscover/blob/main/CONTRIBUTING.md#making-a-pull-request)
- [X] I've updated or added any relevant docstrings following the syntax described in the
- [X] If this PR adds a new feature, I've added tests that sufficiently cover my new functionality.

## After submitting


adam2392:
Hi @francescomontagna wow thanks for this PR! 

I'll take some time over the next week or so to review it. 

In the meantime, WDYT about adding an example or a few examples that demonstrate how to use these methods and when these methods are useful? These are useful to guide users that are not familiar with the methods (e.g. me too :))

You can look at the existing examples/ directory for the format and high level layout.

francescomontagna:
> Hi @francescomontagna wow thanks for this PR!
> 
> I'll take some time over the next week or so to review it.
> 
> In the meantime, WDYT about adding an example or a few examples that demonstrate how to use these methods and when these methods are useful? These are useful to guide users that are not familiar with the methods (e.g. me too :))
> 
> You can look at the existing examples/ directory for the format and high level layout.

Hi @adam2392, thanks for your time! Sounds good to me, I am happy to do that

adam2392 (commented):
Thanks for adding a few examples! Also thanks for describing the general flow of the algorithms in the example too.

I might not fully understand the methods, but they do seem similar enough where we can combine them into one example, where we explain the differences among the different algorithms along with the general workflow. This will simplify maintenance and communication to the user, who will peruse this one example to understand which algorithm of the three to use (if any) for their particular use case. WDYT?

Something like:

```
# <insert: this is the general workflow of the algorithms>
# 
# DAS, NOGAM and SCORE are three proposed algorithms from <insert references>
# that differ in a few ways:
# - <list differences>
```

If there's anything else, feel free to modify/add as you see fit. Will review the code structure next week.

Thanks!

dodiscover/metrics.py
```diff
@@ -146,3 +145,29 @@ def structure_hamming_dist(
         diff = diff + diff.T
         diff[diff > 1] = 1  # Ignoring the double edges.
         return np.sum(diff) / 2
+
+
+def toporder_divergence(A: NDArray, order: List[int]):
```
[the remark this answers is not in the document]
adam2392: This makes it more explicit that this is a metric that operates on an "array-like" structure, whereas the other implemented metrics currently operate on a "networkx-like" structure.

francescomontagna:
> Thanks for adding a few examples! Also thanks for describing the general flow of the algorithms in the example too.
> 
> I might not fully understand the methods, but they do seem similar enough where we can combine them into one example, where we explain the differences among the different algorithms along with the general workflow. This will simplify maintenance and communication to the user, who will peruse this one example to understand which algorithm of the three to use (if any) for their particular use case. WDYT?
> 
> Something like:
> 
> ```
> # <insert: this is the general workflow of the algorithms>
> # 
> # DAS, NOGAM and SCORE are three proposed algorithms from <insert references>
> # that differ in a few ways:
> # - <list differences>
> ```
> 
> If there's anything else, feel free to modify/add as you see fit. Will review the code structure next week.
> 
> Thanks!

Yes, no problems on that side! I will squash the examples into one, it actually make sense as they follow very close logic.

Concerning other things to do, I have a couple of questions:
- On the `CONTRIBUTING.md` file it is asked to update the Changelog file. 
    > And finally, please update the [CHANGELOG](https://github.com/py-why/dodiscover/docs/whats_new.rst) with notes on your contribution in the "Unreleased" section at the top. 

    but I fear that  [CHANGELOG](https://github.com/py-why/dodiscover/docs/whats_new.rst) is broken. Also, in `https://github.com/py-why/dodiscover/doc/v0.1.rst` I did not find any "Unreleased" section. 
- The unit tests keep failing due to an import error in `tests/unit_tests/replearning/test_gin.py`
    `__________ ERROR collecting tests/unit_tests/replearning/test_gin.py ___________
ImportError while importing test module '/home/runner/work/dodiscover/dodiscover/tests/unit_tests/replearning/test_gin.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/unit_tests/replearning/test_gin.py:8: in <module>
    from dodiscover.replearning.gin import GIN
dodiscover/replearning/__init__.py:1: in <module>
    from .gin import GIN
dodiscover/replearning/gin.py:6: in <module>
    from pywhy_graphs.array.export import clearn_arr_to_graph
E   ModuleNotFoundError: No module named 'pywhy_graphs.array.export'`

    Should I worry about this? I am not sure were it is from.

adam2392:
>   but I fear that  [CHANGELOG](https://github.com/py-why/dodiscover/docs/whats_new.rst) is broken. Also, in `https://github.com/py-why/dodiscover/doc/v0.1.rst` I did not find any "Unreleased" section.

You'll want to check `https://github.com/py-why/dodiscover/doc/whats_new/v0.1.rst`. I'll push a fix to the documentation

> * The unit tests keep failing due to an import error in `tests/unit_tests/replearning/test_gin.py`
>   `__________ ERROR collecting tests/unit_tests/replearning/test_gin.py ___________ ImportError while importing test module '/home/runner/work/dodiscover/dodiscover/tests/unit_tests/replearning/test_gin.py'. Hint: make sure your test modules/packages have valid Python names. Traceback: 

Hmm not sure, but perhaps you can add `__init__.py` into the replearning/ test folder. I can't reproduce the error on my machine, so maybe you're using a diff version of Python.

> dodiscover/replearning/gin.py:6: in <module> from pywhy_graphs.array.export import clearn_arr_to_graph E   ModuleNotFoundError: No module named 'pywhy_graphs.array.export'`

Yes this is due to a minor API change in `pywhy-graphs`, so there's a pending PR to fix this. I wouldn't worry about this. I'll try to get it fixed asap, so you can merge in changes from `main` branch.

adam2392:
FYI, that error you raised should be fixed now in main. If you merge in changes from main it should be fine now. @francescomontagna hope that helps!

adam2392:
Xref: https://output.circle-artifacts.com/output/job/e8cf4bf6-34e0-4106-b7d0-90674f9478ea/artifacts/0/dev/auto_examples/plot_score_alg.html#sphx-glr-auto-examples-plot-score-alg-py

dodiscover/metrics.py
```diff
@@ -146,3 +145,29 @@ def structure_hamming_dist(
...
+        A topological ordering on the nodes of the graph.
+
+    Return
+    ------
+    err : int
+        Sum of the number of edges of A not admitted by the given order.
+    """
+    err = 0
+    for i in range(len(order)):
+        err += A[order[i + 1 :], order[i]].sum()
```
robertness, line 172: Suggest more descriptive names for people who are new to the idea

dodiscover/metrics.py
```diff
@@ -146,3 +145,29 @@ def structure_hamming_dist(
         diff = diff + diff.T
         diff[diff > 1] = 1  # Ignoring the double edges.
         return np.sum(diff) / 2
+
+
+def toporder_divergence_arr(A: NDArray, order: List[int]):
```
robertness, line 150: output type?

dodiscover/metrics.py
robertness: maybe something like:

```
from numpy.typing import NDArray
from typing import List

def toporder_divergence_arr(A: NDArray, order: List[int]) -> int:
    if len(order) != A.shape[0] or A.shape[0] != A.shape[1]:
        raise ValueError("The dimensions of the adjacency matrix and the order list do not match.")
        
    total_edges_not_admitted = 0
    for i in range(len(order)):
        total_edges_not_admitted += A[order[i + 1 :], order[i]].sum()
    return total_edges_not_admitted

```

robertness:
Hi @francescomontagna, thanks so much for this contribution!

It may help to break this PR up into chunks so we have less of a bar to get it merged.  But I'll follow @adam2392 's lead.

Perhaps also you might consider giving an informal presentation on your work the PyWhy weekly community meeting?

dodiscover/toporder/_base.py
```diff
@@ -0,0 +1,602 @@
...
+        A_dense : np.ndarray
+            d x d dense adjacency matrix to be pruned.
+
+        Return
+        ------
+        A : np.ndarray
+            The pruned adjacency matrix output of the causal discovery algorithm.
+        """
+        _, d = X.shape
+        # X = X.detach().cpu()
```
robertness, line 367: remove commented out code

dodiscover/toporder/cam.py
```diff
@@ -0,0 +1,236 @@
...
+        # Score gain for each edge i -> j
+        for i in range(d):
+            for j in range(d):
+                if G_excluded.has_edge(i, j):
+                    score_gains[i, j] = self.inf
+                elif score_gains[i, j] != self.inf:
+                    gam = self._fit_model(X[:, i].reshape(-1, 1), X[:, j].reshape(-1, 1))
+                    residuals = X[:, j] - gam.predict(
+                        X[:, i].reshape(-1, 1)
+                    )  # Overfitting? Also, recomputing gam.predict(•) is likely inefficient
```
robertness, line 231: Remove comments not meant for other developers

dodiscover/toporder/cam.py
```diff
@@ -0,0 +1,236 @@
...
+                    gain = (
+                        -np.log(np.var(residuals)) - init_score[j]
+                    )  # np.var(residuals) different from 1/n * np.sum(np.square(residuals))...
```
robertness, line 234: Is this a problem? Has it been addressed?

adam2392:
> It may help to break this PR up into chunks so we have less of a bar to get it merged. But I'll follow @adam2392 's lead.

I'll say up to @francescomontagna since this is just one big PR related to topological order learning. If you're interested in investing some time to develop better OSS skills and practices, then this might be useful for both you and us. The cons w/ the big PRs are that we will most likely run through many review iterations to develop understanding from the maintainers side. OTOH if you think it might be quite tough to break up the PR, then we can stick w/ reviewing this one.

> Perhaps also you might consider giving an informal presentation on your work the PyWhy weekly community meeting?

+1 would be useful if you're willing and might be exciting to get involved in the community :)

dodiscover/toporder/utils.py
```diff
@@ -0,0 +1,49 @@
+from typing import List
+
+import numpy as np
+from numpy.typing import NDArray
+
+
+def full_DAG(top_order: List[int]) -> NDArray:
```
adam2392, line 7:
```suggestion
def full_dag(top_order: List[int]) -> NDArray:
```

Not a big fan of mixing upper-lower case unless it is a class

dodiscover/toporder/utils.py
```diff
@@ -0,0 +1,49 @@
...
+        The fully connected adjacency matrix encoding the input topological order.
+    """
+    d = len(top_order)
+    A = np.zeros((d, d))
+    for i, var in enumerate(top_order):
+        A[var, top_order[i + 1 :]] = 1
+    return A
+
+
+def fullAdj2Order(A: NDArray):
```
adam2392, line 27:
```suggestion
def full_adj_to_order(A: NDArray):
```

Follow snake-casing pattern in general

pyproject.toml
```diff
@@ -39,7 +39,7 @@ include = ['dodiscover*']
 exclude = ['*tests', '*testdata']
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.11"
+python = ">=3.8.1,<3.11"
```
adam2392, line 42: Is this change necessary?

examples/plot_score_alg.py
```diff
@@ -0,0 +1,189 @@
...
+NoGAM :footcite:`Montagna2023b` algorithms perform causal discovery in a two steps procedure.
+Given observational i.i.d. data from an Additive Noise Model without
+latent confounders, first the method estimates a topological ordering of the causal variables.
+This partial oredring can be represented as a fully connected graph, where every node has an
+incoming edge from all its predecessors in the ordering. Second, the resulting fully connected
+DAG is pruned by some variable selection procedure.
+Note that knowing the topological ordering is already sufficient for estimating causal effects.
+Nevertheless, the pruning step is justified by the fact that operating with a sparser graph is
+statistically more efficient.
+
```
adam2392, line 22: It would be helpful to have some brief description as well here about how one estimates the topological order. Then have a brief description on generally what pruning involves. The previous paragraphs are useful because it is now clear to me that what I am observing are a combination of methods that differ in one of the two step procedures.

examples/plot_score_alg.py
```diff
@@ -0,0 +1,189 @@
...
+# Given the inferred topological order, SCORE prunes the graph by with all
+# edges admitted by such ordering, by doing sparse regression to choose the
+# relevant variables. Variable selection is done by thresholding on the
+# p-values of the coefficients associated to the potential parents of a node.
+# For instance, consider a graph G with 3 vertices V = {1, 2, 3}. For
+# simplicity let the topological order be trivial, i.e. {1, 2, 3}. The unique
+# fully connected adjacency matrix compatible with such ordering is the upper
+# triangular matrix np.triu(np.ones((3, 3)), k=1) with all ones above the
+# diagonal.
+score = SCORE()  # or DAS() or NoGAM()
```
adam2392, line 168: Or CAM()?

examples/plot_score_alg.py
```diff
@@ -0,0 +1,189 @@
...
+# %%
+# Run structure learning algorithm
+# --------------------------------
+# Now we are ready to run the SCORE algorithm. The methods performs inference
+# in two phases. First it estimates the topological order of the nodes in the
+# graphs. This is done iteratively according to the following procedure
+# 1. SCORE estimates the Hessian of log p(x), with p(x) joint distribution of
+# nodes in the graph
+# 2. Let H := Hessian(log p(x)). SCORE selects a leaf in the graph by finding
+# the diagonal term of H with minimum variance,
```
adam2392, line 155: For a user, it might be helpful to define what a leaf in a graph is first

adam2392 (changes requested):
Can you add relevant entries related to the new metric and a new section on topological ordering methods in the `doc/api.rst` file?

Xref: https://output.circle-artifacts.com/output/job/e8cf4bf6-34e0-4106-b7d0-90674f9478ea/artifacts/0/dev/api.html

Overall, I took an initial pass on the codebase. There are a few areas that could use some work and explaining. The comments applies to pretty much all the files I didn't touch as well. Once those are done, I can take a deeper look to determine if there's any redundancy in the code.

Take a crack at the changes and let us know if you have any issues/questions!

pyproject.toml
francescomontagna: This is requried by pygam 0.9.0

dodiscover/metrics.py
francescomontagna: I rewrote a bit the docstring explaining that the error introduced by the topological ordering is a lower bound on the false nagatives (i.e. if order does not allow to recover an edge, this will for sure count as false negative in structural hamming distance), and went for 
```
def toporder_divergence_arr(A: NDArray, order: List[int]) -> int:
    if len(order) != A.shape[0] or A.shape[0] != A.shape[1]:
        raise ValueError("The dimensions of the adjacency matrix and the order list do not match.")
        
    false_negatives_from_order = 0
    for i in range(len(order)):
        false_negatives_from_order += A[order[i + 1 :], order[i]].sum()
    return false_negatives_from_order
```

dodiscover/toporder/das.py
```diff
@@ -0,0 +1,127 @@
...
+        max_parents: int = 20,
+    ):
+        super().__init__(
+            eta_G, eta_H, cam_cutoff, n_splines, splines_degree, estimate_variance=True, pns=False
+        )
+        self.min_parents = min_parents
+        self.max_parents = max_parents
+        self.das_cutoff = das_cutoff
+
+    def prune(self, X: NDArray, A_dense: NDArray) -> NDArray:
```
robertness, line 66: You don't access `self` here.  Make it a static method?

dodiscover/toporder/das.py
```diff
@@ -0,0 +1,127 @@
...
+                            hess_l[:, node],
+                            hess_l[:, argmin],
+                            alternative="greater",
+                            equal_var=False,
+                        )
+                        if pval < self.das_cutoff:
+                            parents.append(remaining_nodes[node])
+
+            A_das[parents, leaf] = 1
+            del remaining_nodes[l_index]
```
robertness, line 125: nit: pop method better?

dodiscover/toporder/nogam.py
```diff
@@ -0,0 +1,181 @@
...
+
+        If self.do_pns = True or self.do_pns is None and number of nodes >= 20, then
+        Preliminary Neighbors Search :footcite:`Buhlmann2013` is applied before CAM pruning.
+        """
+        d = A_dense.shape[0]
+        if (self.do_pns) or (self.do_pns is None and d > 20):
+            A_dense = self.pns(A_dense, X=X)
+        return super().prune(X, A_dense)
+
+    def _mse(self, X: NDArray, Y: NDArray) -> List[float]:
```
robertness, line 127:
```suggestion
    def _mse(self, X: NDArray, Y: NDArray) -> List[float]:
    _, d = Y.shape
    err = [
        np.mean((Y[:, col] - cross_val_predict(self._create_kernel_ridge(), X[:, col].reshape(-1, 1), Y[:, col], cv=self.n_crossval))**2).item()
        for col in range(d)
    ]
    return err

```

dodiscover/toporder/nogam.py
```diff
@@ -0,0 +1,181 @@
...
+            response = Y[:, col]
+            explanatory = X[:, col].reshape(-1, 1)
+            regr = KernelRidge(kernel="rbf", gamma=self.ridge_alpha, alpha=self.ridge_alpha)
+            pred = cross_val_predict(regr, explanatory, response, cv=self.n_crossval)
+            res = response - pred
+            mse = (res**2).mean().item()  # rmse = mse / (pred**2).mean().item() # relative MSE
+            err.append(mse)
+        return err
+
+    def _estimate_residuals(self, X: NDArray) -> NDArray:
```
robertness, line 156:
```suggestion
    def _estimate_residuals(self, X: NDArray) -> NDArray:
    R = [
        X[:, i] - cross_val_predict(self._create_kernel_ridge(), np.hstack([X[:, 0:i], X[:, i + 1 :]]), X[:, i], cv=self.n_crossval)
        for i in range(X.shape[1])
    ]
    return np.vstack(R).transpose()
```

dodiscover/toporder/cam.py
francescomontagna: - For comment 1, part 1 `# Overfitting?`: overfitting should be accounted for by the smoothing parameter lambda that is tuned in `_fit_model()` method. This is analogous to original R implementation of CAM
- For comment 1, part 2: `Also, recomputing gam.predict(•) is likely inefficient` I didn't find a way with `pygam` library to retain predictions after fitting, such that I have to recompute them. This is suboptimal, but the impact is relatively small as in `_fit_model()` we have to recompute predictions several times for the lambda tuning anyway. 
- Comment 2: `# np.var(residuals) different from 1/n * np.sum(np.square(residuals))` is addressed.

tests/unit_tests/toporder/test_score.py
```diff
@@ -0,0 +1,261 @@
...
+def dummy_sample():
+    """
+    Ground truth:  [[0, 0, 0, 0],
+                   [1, 0, 0, 0],
+                   [1, 0, 0, 1],
+                   [1, 0, 0, 0]]
+    """
+    X = np.array(
+        [
+            [-0.839573121, -0.277701505, -0.430597219, -0.060861940],
```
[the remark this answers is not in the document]
francescomontagna: These data are synthetically generated from code that I usually use for my research. I think that I would avoid adding that here since it requires some additional dependencies - unless you think it's fine, I don't have strong opinions on this. I can try generate similar data with the AdditiveNoiseModel class from the dowhy.gcm.fcms module, what do you think @adam2392?

dodiscover/toporder/das.py
francescomontagna: Thanks for pointing that. Actually I was creating a `_SteinMixin` instance in the method, even though it is an ancestor of `DAS`. So now I fixed that by accessing `_SteinMixin` methods using `self`

tests/unit_tests/toporder/test_score.py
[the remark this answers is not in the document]
adam2392: Re adding your code: Yes prolly best to avoid adding here, unless they are simple? We could have them in the test file as private data-generating functions.

Re try generating similar data: If it's easy to do, that would be great, just for making the unit-tests a bit more readable/simplified. You can also demonstrate this in the example then, which would be informative.

pyproject.toml
adam2392: Feel free to resolve convos if they are addressed.

examples/plot_score_alg.py
adam2392: Yeah that would work!

dodiscover/toporder/_base.py
```diff
@@ -0,0 +1,602 @@
...
+        Fit GAM on `X` and `y`.
+
+        Parameters
+        ----------
+        X : np.ndarray
+            exogenous variables.
+        y : np.ndarray
+            endogenous variables.
+        formula : TermList
+            List of the additive splines for the GAM fitting formula.
```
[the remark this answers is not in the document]
francescomontagna: Not there, I have to remove it!

dodiscover/toporder/_base.py
```diff
@@ -0,0 +1,602 @@
...
+            estimator of the score function
+        c : int
+            index of the column of interest
+        eta: float
+            regularization parameter for ridge regression in Stein
+            hessian estimator
+        K : np.ndarray
+            Gaussian kernel evaluated at X
+        s : float
+            Width of the Gaussian kernel
```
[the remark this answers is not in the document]
francescomontagna: It is computed inside the function. This is to avoid to recompute it unnecessarily when it is already known. I will add this info in the docstring.

adam2392:
To resolve circleCI errors: Numpydoc wants `Returns`, `Parameters` as section headers, not `Return` and `Parameter`.

adam2392 (commented):
Thanks for the hard work @francescomontagna! It is coming together. 

If you want to debug the circleCI doc errors locally, you can run `poetry run poe build_docs_noplot`, or `build_docs` to build the docs w/o and w/ plotting the examples. 

Another high-level comment I have is to ask if you can add some intuition to the hyper parameters of each algorithm, so that it can inform users on how to select them. Some strategies:

1. Include some blurbs in the `Notes` section of the docstrings of the algorithms
2. Include an additional module, similar to `constraint_causal_discovery.rst`, which sort of gives a more user-friendly introduction to these topological ordering methods.
3. some combination of the above

We can also include this in a downstream PR.

francescomontagna:
> To resolve circleCI errors: Numpydoc wants `Returns`, `Parameters` as section headers, not `Return` and `Parameter`.

Thanks, I missed that! Actually I was focusing on the fact that while building documentation I also get the following circleCI error [https://app.circleci.com/pipelines/github/py-why/dodiscover/1081/workflows/515d2e97-57a1-4cf3-ba07-2648c75052e3/jobs/1671?invite=true#step-120-234](https://app.circleci.com/pipelines/github/py-why/dodiscover/1081/workflows/515d2e97-57a1-4cf3-ba07-2648c75052e3/jobs/1671?invite=true#step-120-234)

This is despite the fact that all the references I point to are actually in the `references.rst` file. Do you have any thoughts on how to fix that, should I worry about it?

tests/unit_tests/toporder/test_score.py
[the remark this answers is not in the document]
francescomontagna: In the end I opted for writing some data generator function `dummy_sample` in `dodiscover.toporder.utils.py` that takes as input a random seed.

dodiscover/metrics.py
```diff
@@ -146,3 +145,42 @@ def structure_hamming_dist(
...
+    If the topological ordering is compatible with the graph ground truth,
+    the divergence is equal to 0. In the worst case of completely reversed
+    ordering, toporder_divergence is equals to P, the number of edges (positives)
+    in the ground truth graph.
+    Note that the divergence defines a lower bound for the Structural Hamming Distance.
+
+    Parameters
+    ----------
+    true_graph : NetworkxGraph
+        Input adjacency matrix representation of a directed acyclic graphs.
```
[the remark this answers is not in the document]
francescomontagna: Yes, I will add that to the docstring

francescomontagna:
Hi @adam2392 , thanks for the patience in the review process. I should have addressed all of the suggestions in the comments. At this point, as proposed by @robertness , if you think that it could be useful then we can consider to break the pull request in smaller chunks. Let me know!

adam2392:
Do you mind going thru all the convos that are resolved and resolve them. And leave the ones that are still to be discussed open? 

It makes the reviewing process a bit easier that way. Thanks! Taking a look now.

dodiscover/toporder/nogam.py
```diff
@@ -0,0 +1,209 @@
...
+                np.hstack([X[:, 0:i], X[:, i + 1 :]]),
+                X[:, i],
+                cv=self.n_crossval,
+            )
+            for i in range(X.shape[1])
+        ]
+        return np.vstack(R).transpose()
+
+    def _create_kernel_ridge(self):
+        return KernelRidge(kernel="rbf", gamma=self.ridge_alpha, alpha=self.ridge_alpha)
```
adam2392, line 209: Just curious. Does it make sense to preserve the KernelRidge estimator as a "fitted" property of the class?

e.g.

`self.kernel_ridge_est_ : KernelRidge`

Would a user ever want/need to query the estimator directly? Would it save efficiency? If not, it's fine. Jw.

dodiscover/toporder/utils.py
```diff
@@ -0,0 +1,119 @@
...
+    A : np.ndarray
+        The fully connected adjacency matrix encoding the input topological order.
+
+    Returns
+    -------
+    top_order : List[int]
+        Topological order encoding of the input fully connected adjacency matrix.
+    """
+    order = list(A.sum(axis=1).argsort())
+    order.reverse()
```
adam2392, line 43: why is reverse necessary? Can you add an inline comment?

adam2392:
https://output.circle-artifacts.com/output/job/b4c024ba-70aa-4ef3-b38d-ee95cde8b678/artifacts/0/dev/auto_examples/plot_score_alg.html#sphx-glr-auto-examples-plot-score-alg-py

TODO: I will review the example now and leave some comments.

adam2392 (changes requested):
I left some minor final comments and questions for my understanding. We can discuss each of the questions I posed on their threads. Please resolve any convos that you feel are fully addressed. 

On my TODO list is to understand the example you put up. If you end up making changes to the example, please just ping me in case I'm reviewing an old version.

Almost there I think. Thanks!

dodiscover/toporder/das.py
francescomontagna: Addressed

dodiscover/toporder/utils.py
francescomontagna: Reverse is necessary because the topological order is meant to have source first, leaves last (and thus I implemented and used it according to this convention). I will add a comment on that

dodiscover/toporder/nogam.py
francescomontagna: The regressor is created multiple times, each of which it is fitted on different data - this happens in the `self.top_order` method when calling `self._estimate_residuals` and `self._mse`.
The `self._create_kernel_ridge` is then just a shortcut that makes easy to regenerate a model with the same parameters.

dodiscover/toporder/_base.py
```diff
@@ -0,0 +1,629 @@
...
+        splines_list : List[Term]
+            List of splines term for the GAM formula.
+            Example: [s(0), s(1), s(2)] where s is a B-spline Term from pyGAM.
+            The equivalent R formula would be "s(0) + s(1) + s(2)", while the y target
+            is provided directly at gam.fit() call
+
+        Returns
+        -------
+        terms : TermList
+            formula of the type requested by pyGAM Generalized Additive Models class.
```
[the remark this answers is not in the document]
francescomontagna: Yes, I will add it in the `Notes`

dodiscover/toporder/_base.py
```diff
@@ -0,0 +1,629 @@
...
+
+        Parameters
+        ----------
+        X_diff : np.ndarray of shape (n_samples, n_samples, n_nodes)
+            Matrix of the difference between samples.
+        """
+        D = np.linalg.norm(X_diff, axis=2)
+        s = np.median(D.flatten())
+        return s
+
```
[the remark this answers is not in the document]
adam2392: Yep! We can sub it out when the time comes.

dodiscover/toporder/nogam.py
adam2392: Ah I see, so the regressor is in some sense a "private attribute" and will never need to be seen by the user. Is that correct?

francescomontagna: Yes

dodiscover/toporder/_base.py
```diff
@@ -0,0 +1,629 @@
...
+        return A
+
+    def _dag_check_included_edges(self) -> None:
+        """Check that the edges included in `self.context` does not violate DAG condition."""
+        is_dag = nx.is_directed_acyclic_graph(self.context.included_edges)
+        if nx.is_empty(self.context.included_edges):
+            is_dag = True
+        if not is_dag:
+            raise ValueError("Edges included in the graph violate the acyclicity condition!")
```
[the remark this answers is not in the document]
francescomontagna: Yes, and it works a bit differently for CAM with respect to the other methods.

- CAM infers the topological ordering by explicitly building a fully connected DAG one edge at the time, since this has a 1:1 mapping with the ordering. While inferring the topological order it uses a binary matrix `directed_paths` of shape `(num,_nodes, num_nodes)` that stores directed paths  between nodes, i.e. `directed_paths[i, j]=1` means that there is a directed path from node `i` to node `j`. In such case, it will simply forbid edges that allow for a cyclic path from `j` to `i` to be added. So, when an edge is included as an external constraint, I simply update the `directed_paths` matrix accordingly
- The other methods infer the topological order vector directly, one leaf at the time (rather then building a fully connected matrix). So, let's say a  user includes `i -> j` as an edge: when selecting leaf nodes during the ordering, I forbid `i` to be picked as a leaf before `j`.  

I hope this clarifies a bit, feel free to ask more.

dodiscover/toporder/_base.py
```diff
@@ -0,0 +1,629 @@
...
+        nablaK : np.ndarray of shape (n_samples, n_nodes)
+            <nabla, K> dot product.
+        """
+        nablaK = None
+        if s is None:
+            s = self._kernel_width(X_diff)
+        K = np.exp(-np.linalg.norm(X_diff, axis=2) ** 2 / (2 * s**2)) / s
+        if evaluate_nabla:
+            nablaK = -np.einsum("kij,ik->kj", X_diff, K) / s**2
+        return K, nablaK
```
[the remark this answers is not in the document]
francescomontagna: Do yoy mean the computation of K could be done directly with some scikit-learn function, or you mean that the whole function coud be replaced by some existing implementation?

dodiscover/metrics.py
```diff
@@ -146,3 +145,29 @@ def structure_hamming_dist(
...
+
+    Given an adjacency matrix A and a topological ordering on its entries,
+    the topological ordering divergence (Rolland et al. (2022)) counts
+    the number of edges of A that can not be recovered due to the choice of
+    the topological ordering.
+
+    Parameters
+    ----------
+    A : NDArray
+        Input adjacency matrix representation of a directed acyclic graphs.
```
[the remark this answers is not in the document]
francescomontagna: I changed the input type of `A` to be a `NetworkxGraph`, and I specified in the docstring that the input must be a DAG - also adding a check that raise an error if the input is not directed acyclic

dodiscover/toporder/_base.py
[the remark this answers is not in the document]
francescomontagna: I will make all methods private except for  `fit`, `top_order`, `prune`. I think that `fit` and `top_order` must be public. For `prune`, that is more arguable: is pruning-only an actual use case?  In that case, one could think of extending the usage of `Context` for adding constraints to the topological order of the graph, which I would nevertheless leave to a downstream PR, if we think that it is a good idea.
Anyway, at the moment I need to use `prune` outside of the class for unit testing, which is why I would consider leaving the method as public.

examples/plot_score_alg.py
francescomontagna: I added the definition "inline" when describing SCORE. If you think that the example is unclear, I will try to be more descriptive on how SCORE finds the ordering, also being more explicit on leaf node definition.

dodiscover/toporder/_base.py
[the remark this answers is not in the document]
adam2392: > I will make all methods private except for `fit`, `top_order`, `prune`. I think that `fit` and `top_order` must be public. For `prune`, that is more arguable: is pruning-only an actual use case? In that case, one could think of extending the usage of `Context` for adding constraints to the topological order of the graph, which I would nevertheless leave to a downstream PR, if we think that it is a good idea.

Yes that would be very cool, so how "loose" is the `prune` method? Can we easily refactor it out of the topological class into some type of `Mixin` class that can be used in other modules?

dodiscover/toporder/_base.py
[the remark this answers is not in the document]
adam2392: I would add this into the `Notes` section of the class to clarify what sort of "priori" knowledge is supported and perhaps also worth mentioning what isn't possible to encode (if any). In the actual `plot_score_alg.py` example, if there's an easy way to demonstrate this usage of apriori knowledge, or having a very very short 2nd example that just demonstrates this, then that would be great!

For context, we're trying to make sure apriori knowledge is usable when it is provided, but the challenge is being very explicit about this to users, otw it's not clear when/where apriori causal knowledge is usable.

tests/unit_tests/toporder/test_nogam.py
```diff
@@ -0,0 +1,118 @@
...
+
+from dodiscover import make_context
+from dodiscover.metrics import structure_hamming_dist, toporder_divergence
+from dodiscover.toporder.nogam import NoGAM
+from dodiscover.toporder.utils import dummy_groundtruth, dummy_sample, full_dag, orders_consistency
+
+
+@pytest.fixture
+def seed():
+    return 42
```
adam2392, line 13: You can just set seed in the test file as `seed = 42` or something. Unless there's a need to run these unit tests over multiple seeds?

examples/plot_score_alg.py
```diff
@@ -0,0 +1,210 @@
...
+# that may be used in downstream structure learning algorithms.
+context = make_context().variables(data=data).build()
+
+# Alternatively, one could say specify some fixed edges.
+# Note that when specifying fixed edges, the resulting graph that is
+# learned is not necessarily a "pure CPDAG". In that, there are more
+# constraints than just the conditional independences. Therefore, one
+# should use caution when specifying fixed edges if they are interested
+# in leveraging ID or estimation algorithms that assume the learned
+# structure is a "pure CPDAG".
```
adam2392, line 161: Do you learn a CPDAG?

examples/plot_score_alg.py
```diff
@@ -0,0 +1,210 @@
...
+
+# .. code-block::Python
+#   included_edges = nx.Graph([('x', 'y')])
+#   context = make_context().edges(include=included_edges).build()
```
adam2392, line 165: Just to clarify, would this be "valid" apriori information that the methods can use? E.g. see https://github.com/py-why/dodiscover/pull/129/files#r1179228315

examples/plot_score_alg.py
```diff
@@ -0,0 +1,210 @@
...
+
+# %%
+# Run structure learning algorithm
+# --------------------------------
+# Now we are ready to run the SCORE algorithm. The method performs inference
+# in two phases. First it estimates the topological order of the nodes in the
+# graphs. This is done iteratively according to the following procedure:
+#
+# 1. SCORE estimates the Hessian of the logarithm of :math:`p(x)`,
+# with :math:`p(x)` the joint distribution of the nodes in the graph.
```
adam2392, line 175:
```suggestion
# 1. SCORE estimates the Hessian of the logarithm of :math:`p(V)`,
# where :math:`p(V)` is the joint distribution of the nodes in the graph (``V`` are the nodes in the graph).
```

Is this right? the Hessian is estimated over the joint distribution of all nodes in the graph? So it is a $d \times d$ matrix where d is the number of nodes?

adam2392: I use `V` just to stick with convention, where `x` would be a singleton

examples/plot_score_alg.py
```diff
@@ -0,0 +1,210 @@
...
+#
+# 2. Let `H := Hessian(log p(x))`. SCORE selects a leaf in the graph by finding
```
adam2392, line 177:
```suggestion
# 2. Let `H := Hessian(log p(V))`. SCORE selects a leaf in the graph by finding
```

examples/plot_score_alg.py
adam2392: Node w/o children is clear to me now.

examples/plot_score_alg.py
```diff
@@ -0,0 +1,210 @@
...
+
+# `score_full_dag.png` visualizes the fully connected DAG representation of
+# the inferred topological ordering.
+# `score_dag.png` visualizes the fully connected DAG after pruning with
+# sparse regression.
+dot_graph = draw(nx.from_numpy_array(full_dag(order), create_using=nx.DiGraph))
+dot_graph.render(outfile="score_full_dag.png", view=True)
+
+dot_graph = draw(graph)
+dot_graph.render(outfile="score_dag.png", view=True)
```
adam2392, line 210: You may consider adding titles to each graph so it's apparent what one is looking at.

See: https://www.pywhy.org/pywhy-graphs/dev/auto_examples/visualization/draw_and_compare_graphs_with_same_layout.html#sphx-glr-auto-examples-visualization-draw-and-compare-graphs-with-same-layout-py

examples/plot_score_alg.py
adam2392, line 1: I'm going to leave some comments now on the example (ref: https://output.circle-artifacts.com/output/job/e895e207-39c3-4f7b-8d95-b41629ebffeb/artifacts/0/dev/auto_examples/plot_score_alg.html#sphx-glr-auto-examples-plot-score-alg-py)

1. Can you fix the rendering of the bullet point list in the introduction 3rd paragraph?
2. it would be helpful to have a short summary of what you observed with the two final graphs in the end (e.g. you can create a new section just called "Summary".

adam2392 (commented):
Thanks for addressing most of the comments! There's I think 1 or 2 discussion threads, but if you think it's easier to handle in a sep GH issue, feel free to open an issue.

I started leaving some comments about the example. These are mainly points around communication and formatting, so hopefully not too cumbersome.

kunwuz:
Thanks so much! That's a really huge contribution. I haven't finished going through the code in details, but score-based methods are so fascinating and would like to talk more about them if you are interested :)

tests/unit_tests/toporder/test_nogam.py
francescomontagna: No, no need of multiple seeds

examples/plot_score_alg.py
francescomontagna: No it learns a DAG. My bad, I left some part of the PC example unchanged. I will go through it again and fix it

examples/plot_score_alg.py
francescomontagna: Yes, each method is unit tested for its behaviour with included edges: https://github.com/francescomontagna/dodiscover/blob/e858ed3861df7897cc722ffbe6dfbb4d5dbe7198/tests/unit_tests/toporder/test_cam.py#L100

I noticed that I have to change the test name from `test_given_adjacency_when_pruning_then_returns_dag_with_context_included_edges` to `test_given_adjacency_when_fitting_then_returns_dag_with_context_included_edges` since I am testing for the behaviour of the whole fitting procedure, not only the pruning (i.e. I verify that constraints are applied also in the ordering phase)

dodiscover/toporder/_base.py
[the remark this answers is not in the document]
francescomontagna: Ok, I will add to the `Notes` that it is possible to encode, meaning specific edges, vs what is not possible, meaning constraints in the order. As we discussed I would add the latter in a separate PR.

For included edges encoding I can add an additional example, do you reccomend to have a separate script or it is fine to have it on top of `plot_score_alg.py`?

francescomontagna:
> Thanks so much! That's a really huge contribution. I haven't finished going through the code in details, but score function-based methods are so fascinating and would like to talk more about them if you are interested :)

Thank you @kunwuz ! I would be more then happy to have a discussion about it, I worked on the topic quite a bit, also in collaboration with Kun at CMU, I think it would be great to have a chat :)

examples/plot_score_alg.py
francescomontagna: > Is this right? the Hessian is estimated over the joint distribution of all nodes in the graph? So it is a 
 matrix where d is the number of nodes?

Exactly

dodiscover/toporder/_base.py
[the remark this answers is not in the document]
adam2392: Is it okay to have it as a separate script to keep this example nice and concise? We can add a reference link between each examples, so they're easy to find. The separate script can be super short and just go straight to demonstrating what sort of prior knowledge is needed. Then link back to this example for anyone wondering what the methods are.

dodiscover/toporder/_base.py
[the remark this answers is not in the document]
adam2392: Maybe I'm wrong, but https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.rbf_kernel.html#sklearn.metrics.pairwise.rbf_kernel computes fast pairwise RBF kernels, which is what you need (minus the kernel_width computation and nabla portion).

Is `X_diff (n_samples, n_samples, n_nodes)` and is `K` (n_samples, n_samples), or (n_samples, n_samples, n_samples)?

examples/plot_score_alg.py
francescomontagna: Hi @adam2392. Once again sorry for the delay, I was busy with the neurips deadline.
Anyway, I think that the Summary paragraph would be helpful, but I did notice something odd going on in the example that makes the interpretation of the results not immediate: similarly to what is already present in the PC example, I instantiate a graph with names of the variables `x, y, z, w`. But after fitting the SCORE model, the name of the variables are mapped to numbers `0, 1, 2, 3`.

This is not happening for the PC example, I think that something happens during the fitting process in SCORE that loses the information on the variables' names, that are then re-assigned to numbers. Do you have any idea why this might be happening? I didn't find any answer to that, maybe you have this information off the shelf.

dodiscover/toporder/_base.py
```diff
@@ -0,0 +1,631 @@
...
+        self._dag_check_included_edges()
+        self.order_constraints = self._included_edges_order_constraints()
+
+        # Inference of the causal order.
+        A_dense, order = self.top_order(X)
+        self.order_ = order
+
+        # Inference of the causal graph via pruning
+        A = self.prune(X, A_dense)
+        self.graph_ = nx.from_numpy_array(A, create_using=nx.DiGraph)
```
adam2392, line 352: See here, @francescomontagna you are converting stuff to a numpy array, but then throwing away the node labels. 

Good catch on the example. We should add a separate unit-test function to verify that the labels before/after fitting match that of what is fed in (this is prolly something basic that should be present for all algos).

examples/plot_score_alg.py
adam2392: See: https://github.com/py-why/dodiscover/pull/129/files#r1197865009

tests/unit_tests/toporder/test_cam.py
```diff
@@ -0,0 +1,147 @@
...
+    G_dense = dummy_dense()
+    context_builder = make_context()
+    context = context_builder.variables(observed=X.columns).edges(include=G_dense).build()
+    model.fit(X, context)
+    A_included = nx.to_numpy_array(model.graph_)
+    A_dense = nx.to_numpy_array(G_dense)
+    assert np.allclose(A_dense, A_included)
+
+
+def test_given_custom_nodes_labels_when_fitting_then_input_output_labels_are_consistent():
```
francescomontagna, line 128: This is the unit test to check whether the labels are preserved in the inferred output graph

dodiscover/toporder/_base.py
francescomontagna: I added the summary paragraph and fixed the bullet point list.
I also fixed the behaviour causing the loss of information and wrote a unit test accordingly. See e.g. https://github.com/py-why/dodiscover/pull/129/files#r1211564581 . This can be easily re-used for any other method.

dodiscover/toporder/_base.py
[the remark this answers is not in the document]
francescomontagna: I added the example to illustrate how to encode priori knowledge for the order-based methods.
See https://github.com/py-why/dodiscover/pull/129/files#diff-c2f2f9f69d6a4fac3ef5d0b96cfc57eb8769f2d7a238a430d328cbeef6d80a58.

I show how to encode both included and excluded edges and plot the graph to illustrate the effect.

dodiscover/toporder/_base.py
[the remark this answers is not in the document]
francescomontagna: I reformatted a bit this module in order to decouple the `prune` method and the topological orer base class. I ended up creating a separate `CAMPruning` class with all functionalities for the CAM-pruning. This way, the pruning can be used without having to implement the `BaseCAMPruning` class, which I removed instead.

dodiscover/toporder/_base.py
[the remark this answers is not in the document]
francescomontagna: I replaced the kernel computation with the sklearn implementation you suggested. Concerning the dimensions, `K` should be (n_samples, n_samples).

adam2392:
Sorry for the wait @francescomontagna! Waiting to get someone to approve #137 so the CIs can be fixed and we can possibly merge this in soon. I think this is looking good/starting to come together. Maybe a few nits at most imo.

Hoping someone else can also take a look since this is introducing a completely new submodule essentially.

adam2392 (approved):
This LGTM! I'll wait a few days to let others chime in in case they have thoughts.

francescomontagna:
> LGTM



> Sorry for the wait @francescomontagna! Waiting to get someone to approve #137 so the CIs can be fixed and we can possibly merge this in soon. I think this is looking good/starting to come together. Maybe a few nits at most imo.
> 
> Hoping someone else can also take a look since this is introducing a completely new submodule essentially.

Hi Adam! No worries, thanks for the update!

adam2392:
Alright merging in for now! Thanks @francescomontagna. 

In the future, perhaps we can stick w/ smaller PRs, but since this was a body of work, happy that this got in!

Feel free to open a PR that adds to the documentation that explains the topological causal discovery. See `user_guide.rst` and `constraint_causal_discovery.rst` for examples.

Issue #109, opened 2023-02-08
CMI estimate for mixed-data


adam2392:
See: https://www.mdpi.com/1099-4300/24/9/1234

We have existing code for CMI-knn and CMI-classifier based approaches, which can probably be used.

PR #144, opened 2023-09-01
Add GranDAG algorithm

eeulig:

This PR adds the following causal discovery method: [Gradient-Based Neural DAG Learning](https://arxiv.org/abs/1906.02226), Lachapelle et al., 2020.

The code is heavily based on the [author's implementation available on GitHub](https://github.com/kurowasan/GraN-DAG) (licensed under MIT license). I noted this in the docstring of the main class and added a remark for each function directly taken from above repository.

I also compared the SHD on ER graphs to the results reported in the paper (Tab. 1, 2, 8, 9) for the first 10 datasets provided by the authors [here](https://github.com/kurowasan/GraN-DAG/tree/master/data):

|       | ER-10-1 | ER-10-4 | ER-20-1 | ER-20-4 | ER-50-1 | ER-50-4 |
| ----- | ------- | ------- | ------- | ------- | ------- | ------- |
| [Paper](https://arxiv.org/abs/1906.02226) | 1.7 $\pm$ 2.5 | 8.3 $\pm$ 2.8 | 4.0 $\pm$ 3.4 | 45.2 $\pm$ 10.7 | 5.1 $\pm$ 2.8 | 102.6 $\pm$ 21.2 |
| DoDiscover | 1.6 $\pm$ 1.6 | 15.3 $\pm$ 2.7 | 2.8 $\pm$ 1.6 | 36.9 $\pm$ 8.5 | 7.1 $\pm$ 4.6 | 99.5 $\pm$ 19.1 |

Overall, results seem to be consistent, and I assume most differences can be attributed to differences in the different dataset samples.

## Before submitting

## After submitting

### DIFF
--- a/dodiscover/continuous/utils.py
+++ b/dodiscover/continuous/utils.py
@@ -1,0 +1,38 @@
+import torch
+from scipy.linalg import expm
+
+
+class TrExpScipy(torch.autograd.Function):
+    """
+    autograd.Function to compute trace of an exponential of a matrix
+    From the author's GitHub implementation: https://github.com/kurowasan/GraN-DAG.
+    """
+
+    @staticmethod
+    def forward(ctx, input):
+        with torch.no_grad():
+            # send tensor to cpu in numpy format and compute expm using scipy
+            expm_input = expm(input.detach().cpu().numpy())
+            # transform back into a tensor
+            expm_input = torch.as_tensor(expm_input, device=input.device)
+            # save expm_input to use in backward
+            ctx.save_for_backward(expm_input)
+
+            # return the trace
+            return torch.trace(expm_input)
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        with torch.no_grad():
+            (expm_input,) = ctx.saved_tensors
+            return expm_input.t() * grad_output
+
+
+def is_acyclic(adjacency: torch.Tensor) -> bool:
+    """Check if the given adjacency matrix is acyclic"""
+    prod = torch.eye(adjacency.shape[0], dtype=adjacency.dtype, device=adjacency.device)
+    for _ in range(1, adjacency.shape[0] + 1):
+        prod = torch.matmul(adjacency, prod)
+        if torch.trace(prod) != 0:
+            return False
+    return True
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -54,6 +54,16 @@ Constraint-based structure learning
    PsiFCI
 
 
+Continuous optimization-based structure learning
+===================================
+
+.. currentmodule:: dodiscover.continuous
+.. autosummary::
+   :toctree: generated/
+
+   GranDAG
+
+
 Order-based structure learning
 ===================================
 
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -55,6 +55,14 @@
   year    = {2019}
 }
 
+@inproceedings{Lachapelle2020,
+  author     = {S{\'{e}}bastien Lachapelle and Philippe Brouillard and Tristan Deleu and Simon Lacoste{-}Julien},
+  title      = {Gradient-Based Neural {DAG} Learning},
+  booktitle  = {8th International Conference on Learning Representations},
+  year       = {2020},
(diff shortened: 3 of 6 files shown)


adam2392:
Wow thanks for starting this very cool feature as a PR! We are always happy to welcome new contributors. 

We generally have to review through this for a number of iterations to usually get a large chunk of code to a mergable state. This is mainly for maintainability, code robustness and understanding. 

For example see the PR for topological learning in #109

adam2392:
Hi @eeulig wow thanks for this PR that covers an important part of causal discovery!

This is as I understand a non-linear extension of NoTears right? This might go through a number of iterations to get the code to a mergable state. This is mainly for maintainability, code robustness and understanding. For example see the PR for topological methods #129. If you're new to contributing to open source, this will also be a great learning experience in improving your scientific coding chops! Welcome to the community :) 

I'll take some time over the next week or so to review it. In the meantime, WDYT about adding an example or a few examples that demonstrate how to use these methods and when these methods are useful? These are useful to guide users that are not familiar with the methods (e.g. me too :))

You can look at the existing examples/ directory for the format and high level layout.

Summary of what we'll work towards:

1. have a consistent API: You can follow the same API as the topological methods, or FCI. E.g. basically implement `learn_graph(data, context)`. Most other methods can usually be private unless used elsewhere.
2. have unit tests that sufficiently cover the relevant use-cases: I see you have added unit tests, so I will review these to see if there's anything missing
3. have some documentation to explain to users how to use the feature: The best is an example like we have [here](https://github.com/py-why/dodiscover/tree/main/examples)

Lmk if this sounds good to you?

pyproject.toml
```diff
@@ -49,6 +49,7 @@ importlib-resources = { version = "*", python = "<3.10" }
 pywhy-graphs = { git = "https://github.com/py-why/pywhy-graphs.git", branch = 'main', optional = true }
 pygraphviz = { version = "^1.11", optional = true }
 pygam = "^0.9.0"
+torch = "^2.0.1"
```
robertness, line 52: I think we are happy to have a torch dependency

dodiscover/continuous/utils.py
```diff
@@ -0,0 +1,38 @@
...
+            return torch.trace(expm_input)
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        with torch.no_grad():
+            (expm_input,) = ctx.saved_tensors
+            return expm_input.t() * grad_output
+
+
+def is_acyclic(adjacency: torch.Tensor) -> bool:
```
robertness, line 31: @adam2392  maybe the acyclicity check on adjacency matrix should go to pywhy-graph?

adam2392: Yes this is already implemented there

eeulig:
Thanks a lot @adam2392 and @robertness ! I'm happy to look into providing an example similar to the ones in `dodiscover/examples` with some further explanations!

I noticed that some checks were not successful, because some gpu related libraries (`libcudnn.so`, `libcublas.so`) could not be loaded. I think this might be because poetry installs a gpu version of torch instead of the cpu version in the test environment. Do you think this is an issue and we should enforce the cpu version of torch instead?

robertness:
Is this something that could be solved with code that sets the device to cpu unless cuda is available?

adam2392:
> Thanks a lot @adam2392 and @robertness ! I'm happy to look into providing an example similar to the ones in dodiscover/examples with some further explanations!

Great! I will try to find some time next week to start taking a look at this. Since this is an "extension" of notears, I want to see if it is possible to design the code in such a way to allow notears to be implemented modularly in a future PR.

> I noticed that some checks were not successful, because some gpu related libraries (`libcudnn.so`, `libcublas.so`) could not be loaded. I think this might be because poetry installs a gpu version of torch instead of the cpu version in the test environment. Do you think this is an issue and we should enforce the cpu version of torch instead?

Yeah I think testing GPU on CIs is notoroiously difficult, so we can force the CPU version and the tests and examples should be super small-scale for the sake of just allowing it to run.

